Issue #29, opened 2018-06-05
Scrollbars


markusressel:
Hey, awesome library :)

I want to use this to build a Text Editor and want to add scrollbars for both x and y axis.
Is there some kind of listener/callback API that I missed to intercept changes in the current positioning of the ZoomLayout? If so I would be able to implement scrollbars myself and update them even when "flinging" the position.

It would also be cool to have something like this (scrollbars) built in - although it might be out of scope for this library.

Thx for the great work.

natario1:
Thanks!

You should override onUpdate(). And get X and Y from the matrix (or ask the engine). Take a look at the ZoomLayout source code, it's pretty simple.

I think it would be in scope for both ZoomLayout and ZoomImageView so if you work this out feel free to submit a PR. Note, here we would want to let Android draw them using the view callbacks, not draw 'custom' scrollbars onDraw

markusressel:
I'm not sure I can follow you on "let Android draw them using the view callbacks". As far as I can tell neither FrameLayout nor ImageView have built in support for scrollbars? So the only way to show scrollbars would be to add them by hand and update their position using the onUpdate() method.

Do you simply mean to use "Views" (added to the FrameLayout) which android draws instead of manually drawing rectangles in onDraw?

natario1:
It's been a long time but as far as I remember, `View` or `ViewGroup` have builtin support for drawing scrollbars.

But the view must implement a few methods, like `computeVerticalScrollRange` or something like that, and then the base class will draw scrollbars based on the `android:scrollbars` attribute.

So if I remember correctly, you should just implement this 2? 4? methods by using the engine values. A good starting point might be the NestedScrollView source code from Google.

natario1:
Oh and you should call `awakenScrollBars` when change is detected. Basically on all updates.

markusressel:
Thx for the help, now I understand :blush:

I will look into implementing this somewhere around (hopefully) this or next week.

PR #30, opened 2018-06-07
Feature/scrollbars

markusressel:
* implemented scrollbars for both ZoomLayout and ZoomImageView

There is some reordering going on in the xml file, hope you don't mind.

This is a PR for #29

### DIFF
--- a/library/src/main/java/com/otaliastudios/zoom/ZoomImageView.java
+++ b/library/src/main/java/com/otaliastudios/zoom/ZoomImageView.java
@@ -4,7 +4,6 @@ import android.annotation.SuppressLint;
 import android.content.Context;
 import android.content.res.TypedArray;
 import android.graphics.Matrix;
-import android.graphics.Rect;
 import android.graphics.RectF;
 import android.graphics.drawable.Drawable;
 import android.support.annotation.AttrRes;
@@ -18,7 +17,7 @@ import android.widget.ImageView;
 
 /**
  * Uses {@link ZoomEngine} to allow zooming and pan events to the inner drawable.
- *
+ * <p>
  * TODO: support padding (from inside ZoomEngine that gets the view)
  */
 @SuppressLint("AppCompatCustomView")
@@ -101,13 +100,6 @@ public class ZoomImageView extends ImageView implements ZoomEngine.Listener, Zoo
     }
 
     @Override
-    protected int computeHorizontalScrollExtent() {
-        Rect localRect = new Rect();
-        getLocalVisibleRect(localRect);
-        return localRect.width();
-    }
-
-    @Override
     protected int computeHorizontalScrollOffset() {
         return (int) (-1 * mEngine.getPanX() * mEngine.getRealZoom());
     }
@@ -115,13 +107,6 @@ public class ZoomImageView extends ImageView implements ZoomEngine.Listener, Zoo
     @Override
     protected int computeHorizontalScrollRange() {
         return (int) (mDrawableRect.width() * mEngine.getRealZoom());
-    }
-
-    @Override
-    protected int computeVerticalScrollExtent() {
-        Rect localRect = new Rect();
-        getLocalVisibleRect(localRect);
-        return localRect.height();
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


natario1:
Man thanks! I will take a look as soon as I can, nice work

markusressel:
Travis is missing the local.properties file which is ignored by git.
Is this "ok"?

natario1:
Yeah you can ignore that, I will fix it later

library/src/main/java/com/otaliastudios/zoom/ZoomImageView.java
```diff
@@ -89,12 +90,50 @@ public boolean onTouchEvent(MotionEvent ev) {
...
 
     @Override
     public void onIdle(ZoomEngine engine) {
     }
 
+    @Override
+    protected int computeHorizontalScrollExtent() {
+        Rect localRect = new Rect();
+        getLocalVisibleRect(localRect);
+        return localRect.width();
```
natario1: Is there any reason why using this rather than `getWidth()` and `getHeight`? It should return the same thing almost always. Also in vertical extent & in ZoomLayout callbacks.

If there is a reason, could you cache this rect in a tempRect field? So we don't create a lot of them.

library/src/main/java/com/otaliastudios/zoom/ZoomLayout.java
```diff
@@ -150,23 +151,66 @@ public void onUpdate(ZoomEngine helper, Matrix matrix) {
...
-            boolean result = super.drawChild(canvas, child, drawingTime);
+            result = super.drawChild(canvas, child, drawingTime);
             canvas.restoreToCount(save);
-            return result;
         } else {
-            return super.drawChild(canvas, child, drawingTime);
+            result = super.drawChild(canvas, child, drawingTime);
         }
+
+        onDrawScrollBars(canvas);
```
natario1: You can remove this and use `setWillNotDraw(false)` in the constructor.

library/src/main/java/com/otaliastudios/zoom/ZoomImageView.java
```diff
@@ -89,12 +90,50 @@ public boolean onTouchEvent(MotionEvent ev) {
     public void onUpdate(ZoomEngine helper, Matrix matrix) {
         mMatrix.set(matrix);
         setImageMatrix(mMatrix);
+
+        if (!awakenScrollBars()) {
+            invalidate();
```
natario1: Any reason for this call? We should not need to invalidate. setImageMatrix will do it.
Same for ZoomLayout.

natario1 (commented):
Just a few comments, looking forward to merge this

library/src/main/java/com/otaliastudios/zoom/ZoomImageView.java
markusressel: The documentation for ```awakenScrollBars()``` states:

```
When the animation is started, this method returns true, 
and false otherwise. If the animation is started, this method 
calls {@link #invalidate()}; in that case 
the caller should not call {@link invalidate()}.
```

So I added this condition. I guess if ```setImageMatrix``` already does it this check is useless.

library/src/main/java/com/otaliastudios/zoom/ZoomImageView.java
markusressel: No, I think it should also work with ```getWidth()``` and ```getHeight()```.
I don't have that much experience with custom views so the "localVisibleRect" method looked like a good fit, but I guess it can be slow.

library/src/main/java/com/otaliastudios/zoom/ZoomImageView.java
natario1: Yeah I think it's better to simply call awakenScrollBars() and ignore the result.
Also, you need to merge the other PR changes to merge this one. Thanks!

markusressel:
I removed the if condition, merged master and added xml attributes for demo ZoomImageView.

natario1 (approved):

natario1:
Thanks @markusressel ! I will release a new version today with your changes.

Issue #16, opened 2018-02-27
Disable Zoom


nikolamk:
Is there any way to temporary pause ( disable ) ZoomLayout  from zoom and pinch? I want ZoomLayout  to stay at paused position until I resume it back  ( children  should keep listening their listeners) 

Thanks for this great component !

natario1:
No. This goes along #15, we should add `zoomEnabled` , `verticalPanEnabled` and `horizontalPanEnabled` , both as XML attributes and methods.

natario1:
This was fixed by #31

Issue #15, opened 2018-02-15
Disable Scrolling (Vertical, Horizontal)


JohnReinel98:
Hey I just want to use pinch zoom, can you add an option which disables the scroll vertically and horizontally? Thanks.

natario1:
Fixed by #31

PR #31, opened 2018-06-07
Feature/disable zoom

markusressel:
* implemented zoomEnabled, horizontalPanEnabled and verticalPanEnabled options

Note that this code only controls zoom and pan input **from the user**. Methods like ```zoomIn()``` or ```panBy()``` will still work even if any of the above mentioned attributes are set to ```false```.

This is a PR for Issue #16 and #15


markusressel:
Well there's a nother travis issue in this one... hope you can fix it easily.

natario1:
Thanks for this one as well 👍 

About the **from the user** part, is this by choice? Do you think it's more useful than disabling it for all uses? <strike>(it probably would be easy to implement by returning false in `setState()`)</strike>.

I'm just asking, I have not thought about this.

natario1:
I have taken a look at the source now. Yeah, let's not block the API usage.

But about implementation in this PR, could you use `setState()` and return false there for scroll / pinch / fling based on the flag? So you can get rid of if branches, 0 velocity and so on. Would it work?

markusressel:
As for the "from the user" part, yes this is intentionally.
I don't think it's a good design to block the programmer from calling things as he should know about the state of his application anyway. 

A more sophisticated implementation could allow disabling things for the user and the programmer separately, but I don't think this really is a necessity.

I think I could use the ```setState()``` method to completely block scrolling and pinching, but to block vertical and horizontal scroll separately I need the "0 velocity" thingy, or am I missing something?

markusressel:
Just to add to the last section:
If the state only has control over full block/no block imho the current solution is the simpler and therefore better one, as the boolean that affects the behaviour is used (accessed) at the exact spots where the behaviour is changed, and doesn't change some "abstract" state that implicitly changes the behaviour.

Imho changing the behaviour partly via state and partly using 0 values that should come from the same "preference" (the enabled flag) would complicate things unnecessarily.

But of course you are welcome to disagree with me on this.

natario1:
No, I agree, I didn't think about the vertical / horizontal issue.

natario1 (approved):
Thanks!

PR #35, opened 2018-06-15
add missing invalidate call

### LINKED ISSUE #34: Scrollbars not rendered when fadeScrollbars="false"
Because of a missing "invalidate()" call (removed in the review of the recently PR) scrollbars are not rendered when the fadeScrollbars attribute is set to false.

This only applies to the ZoomLayout, the ZoomImageView is not affected, apparently because it uses ```setImageMatrix()``` which internally calls ```invalidate()``` on every draw.

I would just follow the documentation and call ```invalidate``` if ```awakenScrollbars()``` returns false.

I will create a PR for this issue shortly.

markusressel:
this fixes #34

### DIFF
--- a/library/src/main/java/com/otaliastudios/zoom/ZoomLayout.java
+++ b/library/src/main/java/com/otaliastudios/zoom/ZoomLayout.java
@@ -159,7 +159,9 @@ public class ZoomLayout extends FrameLayout implements ZoomEngine.Listener, Zoom
             invalidate();
         }
 
-        awakenScrollBars();
+        if (!awakenScrollBars()) {
+            invalidate();
+        }
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


natario1:
What happens if you call just awakenScrollBars() but before all the rest, right after matrix.set()?
I'd like to avoid this extra invalidate call for who is not using scrollbars at all.

markusressel:
Tried it didn't work. We can test though if vertical and/or horizontal scrollbar is enabled if you want.

markusressel:
The problem is awakenScrollbars does nothing if fadeScrollbars is set to false. I don't know why the child views are rendered and the scrollbars are not, but the additional invalidate fixes it. If you know a better solution let me know.

natario1:
Yeah this is confusing, but let's do it this way then 👍thanks for spotting this!

Issue #36, opened 2018-06-16
Convert code to Kotlin


markusressel:
Hey, I really dig this library and would love to see it come to life in Kotlin.
I know you said you would migrate this library to Kotlin in the long run (https://github.com/natario1/ZoomLayout/issues/2).

I would like to take the chance and convert the existing code to a version that is usable for both Kotlin and Java.

natario1:
Sure, let's do it!

markusressel:
Since #38 is now merged this issue is closed. For any kotlin related changes please create a new issue.

Issue #2, opened 2017-09-28
Double tap to zoom


natario1:
It's common to react to double taps to control zoom. This should be a opt-in feature in `ZoomEngine`, enabled by default in `ZoomImageView`.

HardikMaru:
@natario1 

For now, I have just implemented this functionality by my self with custom Gesture Listener class. Here is my code snippet if anyone need it.

```
private GestureDetector gestureDetector; 
private View.OnTouchListener touchListener;
private ZoomImageView selectedImage;
```

then initialize these variables:

```
    gestureDetector = new GestureDetector(mContext, new MyGestureListener());

        touchListener = new View.OnTouchListener() {
            @Override
            public boolean onTouch(View v, MotionEvent event) {
                // pass the events to the gesture detector
                // a return value of true means the detector is handling it
                // a return value of false means the detector didn't
                // recognize the event
                selectedImage = (ZoomImageView) v;
                return gestureDetector.onTouchEvent(event);

            }
        };
```

 then assign your touchlistener to zoomimageview object:

```
    ZoomImageView image = layout.findViewById(R.id.imageViewItemImageSlider);
    image.setOnTouchListener(touchListener);
```

and here is MyGestureListener class:

```
    class MyGestureListener extends GestureDetector.SimpleOnGestureListener {

//        @Override
//        public boolean onDown(MotionEvent event) {
//            Log.d("TAG","onDown: ");
//
//            // don't return false here or else none of the other
//            // gestures will work
//            return true;
//        }
//
//        @Override
//        public boolean onSingleTapConfirmed(MotionEvent e) {
//            Log.i("TAG", "onSingleTapConfirmed: ");
//            return true;
//        }
//
//        @Override
//        public void onLongPress(MotionEvent e) {
//            Log.i("TAG", "onLongPress: ");
//        }

        @Override
        public boolean onDoubleTap(MotionEvent e) {
            Log.i("TAG", "onDoubleTap: ");
            if((selectedImage.getEngine().getZoom() >= 2.75F)) {
                selectedImage.getEngine().zoomTo(1F, true);
            } else if((selectedImage.getEngine().getZoom() < 1F)) {
                selectedImage.getEngine().zoomTo(1F, true);
            }  else {
                selectedImage.getEngine().zoomBy(2F, true);
            }

            return true;
        }

//        @Override
//        public boolean onScroll(MotionEvent e1, MotionEvent e2,
//                                float distanceX, float distanceY) {
[...]

PR #38, opened 2018-06-21
Feature/kotlin

markusressel:
This is a PR for #36

As of now I have tested the basic functionality of the library (as in using it on the emulator) and it seems to work just like the java version.

For this PR I tried to convert the code and leverage delegation and other features of Kotlin to minimize boilerplate. I did **not** try to change any logic to speed up performance or "improve" other things. This PR is only intended to use Kotlin as a way to write the same logic with way less and more readable code that has Null-Safety built in for free.

Some things I was unsure about when fleshing out the details:
* The documentation probably needs a bit of an overhaul, which I didn't tackle yet
* Do you really want to expose the underlying ```ZoomEngine``` with all it's features outside of implementations like ```ZoomLayout``` and ```ZoomImageView```? Previously it was possible to access the ZoomEngine by asking f.ex. ```ZoomLayout```. I think all methods exposed to the outside should be defined in the ZoomApi so I made that field private.
* I used property delegation to give ```ZoomLayout``` and ```ZoomImageView``` users the ability to use the ```ZoomApi``` without the need to redundantly specify the "delegation methods" that simply point to the inner ```engine``` field. To use delegation I had to change the ZoomEngine constructor so it doesn't need ```this``` as a parameter anymore. This was ok for the ```Listener```, but not for the ```container```. Since I didn't want to make the container type ```Nullable``` (```View?```) I initialized it with an empty inline View class. If you know a better solution to this let me know.

I know this is a pretty big change (could it be any bigger?), so I am open to any kind of discussion.

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -1,6 +1,8 @@
 apply plugin: 'com.android.library'
+apply plugin: 'kotlin-android'
 apply plugin: 'com.github.dcendents.android-maven'
 apply plugin: 'com.jfrog.bintray'
+apply plugin: 'org.jetbrains.dokka-android'
 
 // Required by bintray
 // archivesBaseName is required if artifactId is different from gradle module name
@@ -30,6 +32,7 @@ android {
 
 dependencies {
     api "androidx.annotation:annotation:$supportLibVersion"
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
 }
 
 
@@ -133,9 +136,17 @@ task javadocJar(type: Jar, dependsOn: javadoc) {
     from javadoc.destinationDir
 }
 
+task dokkaJar(type: Jar, dependsOn: dokka) {
+    classifier = 'javadoc'
+    from dokka.outputDirectory
+}
+
 artifacts {
-    archives javadocJar
+    archives dokkaJar
     archives sourcesJar
+}
+repositories {
+    mavenCentral()
 }
 
 // export JAVA_HOME=/Library/Java/JavaVirtualMachines/jdk1.8.0_101.jdk/Contents/Home
--- a/library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
+++ b/library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
@@ -1,70 +1,205 @@
-package com.otaliastudios.zoom;
+package com.otaliastudios.zoom
 
 import android.annotation.SuppressLint
 import android.content.Context
 import android.graphics.Matrix
 import android.graphics.RectF
 import android.os.Build
-import android.view.GestureDetector
-import android.view.Gravity
-import android.view.MotionEvent
-import android.view.ScaleGestureDetector
+import android.view.*
+import android.view.animation.AccelerateDecelerateInterpolator
+import android.widget.OverScroller
 import androidx.annotation.IntDef
-import androidx.annotation.NonNull
-import java.lang.annotation.Retention
-import java.lang.annotation.RetentionPolicy
-import java.util.*
+import com.otaliastudios.zoom.ZoomApi.*
 
 
 /**
  * A low level class that listens to touch events and posts zoom and pan updates.
- * The most useful output is a {@link Matrix} that can be used to do pretty much everything,
+ * The most useful output is a [Matrix] that can be used to do pretty much everything,
  * from canvas drawing to View hierarchies translations.
- * <p>
+ *
+ *
  * Users are required to:
  * - Pass the container view in the constructor
- * - Notify the helper of the content size, using {@link #setContentSize(float, float)}
- * - Pass touch events to {@link #onInterceptTouchEvent(MotionEvent)} and {@link #onTouchEvent(MotionEvent)}
- * <p>
- * This class will apply a base transformation to the content, see {@link #setTransformation(int, int)},
+ * - Notify the helper of the content size, using [setContentSize]
+ * - Pass touch events to [onInterceptTouchEvent] and [onTouchEvent]
+ *
+ *
+ * This class will apply a base transformation to the content, see [setTransformation],
  * so that it is laid out initially as we wish.
- * <p>
+ *
+ *
  * When the scaling makes the content smaller than our viewport, the engine will always try
  * to keep the content centered.
  */
-public final class ZoomEngine implements ZoomApi {
+class ZoomEngine
+/**
+ * Constructs an helper instance.
+ *
+ * @param context   a valid context
+ */
+internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener, ZoomApi {
 
-    // TODO add OverScrollCallback and OverPinchCallback.
-    // Should notify the user when the boundaries are reached.
+    /**
+     * Constructs an helper instance.
+     *
+     * @param context   a valid context
+     * @param container the view hosting the zoomable content
+     **/
+    constructor(context: Context, container: View) : this(context) {
+        setContainer(container)
+    }
 
-    // TODO expose friction parameters, use an interpolator.
+    /**
+     * Constructs an helper instance.
+     * Deprecated: use [addListener] to add a listener.
+     *
+     * @param context   a valid context
+     * @param container the view hosting the zoomable content
+     * @param listener a listener for events
+     **/
+    @Deprecated("Use [addListener] to add a listener.",
+            replaceWith = ReplaceWith("constructor(context, container)"))
+    constructor(context: Context, container: View, listener: Listener) : this(context, container) {
+        addListener(listener)
+    }
 
-    // TODO Make public, add API.
-    private static final float DEFAULT_OVERSCROLL_FACTOR = 0.10F;
+    private val mListeners = mutableListOf<Listener>()
+    private var mMatrix = Matrix()
+    private val mOutMatrix = Matrix()
+    @State
+    private var mState = NONE
+    private lateinit var mContainer: View
+    private var mContainerWidth = 0.toFloat()
+    private var mContainerHeight = 0.toFloat()
+    private var mInitialized = false
+    private var mTransformedRect = RectF()
+    private var mContentRect = RectF()
+    private var mMinZoom = 0.8f
+    private var mMinZoomMode = ZoomApi.TYPE_ZOOM
+    private var mMaxZoom = 2.5f
+    private var mMaxZoomMode = ZoomApi.TYPE_ZOOM
 
-    // TODO Make public, add API. Use androidx.Interpolator?
-    private static final Interpolator ANIMATION_INTERPOLATOR = new AccelerateDecelerateInterpolator();
+    /**
+     * Gets the current zoom value, which can be used as a reference when calling
+     * [zoomTo] or [zoomBy].
+     *
+     *
+     * This can be different than the actual scale you get in the matrix, because at startup
+     * we apply a base transformation, see [setTransformation].
+     * All zoom calls, including min zoom and max zoom, refer to this axis, where zoom is set to 1
+     * right after the initial transformation.
+     *
+     * @return the current zoom
+     * @see realZoom
+     */
+    @Zoom
+    @get:Zoom
+    override var zoom = 1f // Not necessarily equal to the matrix scale.
+        private set
+    private var mBaseZoom = 0.toFloat() // mZoom * mBaseZoom matches the matrix scale.
+    private var mTransformation = ZoomApi.TRANSFORMATION_CENTER_INSIDE
+    private var mTransformationGravity = Gravity.CENTER
+    private var mOverScrollHorizontal = true
+    private var mOverScrollVertical = true
+    private var mHorizontalPanEnabled = true
+    private var mVerticalPanEnabled = true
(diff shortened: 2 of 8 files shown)


markusressel:
Since javadoc doesn't support Kotlin (which breaks the build atm) I would recommend to use [Dokka](https://github.com/Kotlin/dokka) instead. It supports Java as well as Kotlin. I will look into this later.

natario1:
Thanks!

I will take a look when I find some time. Meanwhile, the issue with both your bullet points is that they would be breaking changes. I agree that with ZoomApi the engine could be hidden, but someone might be using it, so we don't want to break their code.

Same applies to delegation, let's not change the constructors. You can use an internal constructor without "this" but still provide the old constructor.

I would also make the `container` a `private lateinit var` and create an internal function setContainer() to initialize.

Anyway, thanks, it's cool to see this in Kotlin!
About Dokka & publishing I have spent some hours on this recently, have a working lib here https://github.com/natario1/Elements/blob/master/library/build.gradle

library/build.gradle
```diff
@@ -31,6 +32,7 @@ android {
 
 dependencies {
     api "com.android.support:support-annotations:$supportLibVersion"
+    compile "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
```
natario1: should use `implementation` instead of compile

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -0,0 +1,1208 @@
...
+ */
+class ZoomEngine
+/**
+ * Constructs an helper instance.
+ *
+ * @param context   a valid context
+ * @param container the view hosting the zoomable content
+ * @param listener a listener for events
+ */
+constructor(context: Context, private var container: View, var listener: Listener?)
```
natario1: Can we do it the other way round?

Primary constructor should be `internal constructor(context: Context)`.

Then a secondary constructor will be `public constructor(context: Context, container: View, var listener: Listener?)`. This secondary constructor does two things, calls `this(context)` and calls `setContainer(container)` function.

About the setContainer function, it should be called only once and be `internal`. Something like

```kotlin
private lateinit var mView: View

internal fun setContainer(container: View) {
    mView = container
    // Register onGlobalLayout
}
```

This is to

- delegate constructors the correct way, avoiding object: View(context) {}
- internally keep the `mView` name for the variable, which is more readable than `container` since we also have `mViewWidth` `mViewHeight` and all m* stuff.
- avoid calling addOnGlobalLayoutListener(this) twice (both in the init block and in setContainer function)

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -0,0 +1,1208 @@
...
+        return true
+    }
+
+    /**
+     * Set a container to perform transformations on.
+     * Any previously set container will no longer be modified.
+     *
+     * @param newContainer view
+     */
+    fun setContainer(newContainer: View) {
```
natario1: Already discussed in other comment. & let's not make this public, use internal

library/src/main/java/com/otaliastudios/zoom/ZoomImageView.kt
```diff
@@ -0,0 +1,104 @@
...
+
+/**
+ * Uses [ZoomEngine] to allow zooming and pan events to the inner drawable.
+ *
+ *
+ * TODO: support padding (from inside ZoomEngine that gets the view)
+ */
+@SuppressLint("AppCompatCustomView")
+class ZoomImageView
+@JvmOverloads constructor(context: Context, attrs: AttributeSet? = null, @AttrRes defStyleAttr: Int = 0, val engine: ZoomEngine = ZoomEngine(context))
```
natario1: We don't want the full constructor to be public.

So, the primary constructor should be private. Then we can have 3 secondary constructors that pass the engine instance to the primary & you can use delegation.

library/src/main/java/com/otaliastudios/zoom/ZoomLayout.kt
```diff
@@ -0,0 +1,225 @@
...
+ *
+ * Currently padding to this view / margins to the child view are NOT supported.
+ *
+ *
+ * TODO: support padding (from inside ZoomEngine that gets the view)
+ * TODO: support layout_margin (here)
+ */
+class ZoomLayout
+@JvmOverloads
+constructor(context: Context, attrs: AttributeSet? = null, @AttrRes defStyleAttr: Int = 0, val engine: ZoomEngine = ZoomEngine(context))
```
natario1: See ZoomImageView

library/src/main/java/com/otaliastudios/zoom/ZoomLogger.kt
```diff
@@ -0,0 +1,99 @@
...
+
+    private fun string(messageLevel: Int, vararg ofData: Any): String {
+        var message = ""
+        if (should(messageLevel)) {
+            for (o in ofData) {
+                message += o.toString()
+                message += " "
+            }
+        }
+        return message.trim { it <= ' ' }
```
natario1: I don't know what this does but it creates an object for each Log call right? Let's use simple trim or, better, let's not trim at all..

natario1 (commented):
Sorry for the delay! I've been through some stuff, moved to another country and so on.

markusressel:
No problem I'm pretty busy right now too. Will take a look in approx. 2 weeks when exams are over.

natario1:
Cool!

library/src/main/java/com/otaliastudios/zoom/ZoomLogger.kt
markusressel: This is generated code from the java -> kotlin conversion. I removed the trim.

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
markusressel: done

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
markusressel: done

library/build.gradle
markusressel: done

library/src/main/java/com/otaliastudios/zoom/ZoomImageView.kt
markusressel: I'm not sure if I got what you mean, please take a look :)

library/src/main/java/com/otaliastudios/zoom/ZoomLayout.kt
markusressel: See above

markusressel:
Well it was a little over 2 weeks (shame on me) but I finally made it.
I hope you can still use this to some degree. I saw you already did some work on migrating this to AndroidX and a lot of other stuff too. I did *not* have the time yet to look into documentation but as this is a thing I will need for my own libraries I hope to get around this soon™.

Tadaboody:
Any help needed to close this PR? I'd like this to get merge so I can help with some other issues, namely #2 without it disrupting this PR

natario1:
I am super busy and I have not reviewed changes but I am glad you made it @markusressel .

I don't know how to manage this, I have an open PR at #59 which I am actively working at (there are more commits in my local machine + more coming) and I am in need of pushing that forward. It changes constructors, logic, adds stuff, removes stuff.

Either we wait for #59 and then rewrite this and merge this quickly, or merge this now in a `kotlin` branch. Later improve the `kotlin` branch until it matches java. But it will be hard without a single diff to look at. What do you guys think.

markusressel:
androidx is pretty big and useful as are the other parts of your PR. If it lessens the pressure on you @natario1 I'm happy with reapplying the knowledge of this PR to the version after #59 is merged and apply it to this PR. We just have to find a time where the switch to kotlin is doable and can get merged before making huge architectural changes again.

natario1:
Sorry for the delay. I will merge #59 now so @markusressel if you still want to go with this kotlin job, that would be cool.

When you start we could add a little header to README to block new contributions for a while.
If you would like to have write access here, take care about issues or whatever else, let me know. You have already set up some well thought PRs and seem to be interested in the project. I don't think the amount of work is going to increase, but my time decreases :-)

I think it makes sense (like you did here?) to leave the demo app in java, so we can ensure that the API surface did not change from there.

markusressel:
Great! I hope I can find some extra time this week(end) and get this language port out of the way soon.

Well my time isn't exactly increasing either (hence the delay xD) but since I already know pretty much all of the codebase I may help where I can.

Yes, my intention also was to leave the demo app in java to ensure backwards compatibility is still there.

markusressel:
There are some things that need to be addressed before this can get merged:

- [x] there is still some bug with overpinching - **fixed**
- [x] documentation is not yet generated for kotlin code.

markusressel:
@natario1 Ok so from my point of view this is a state that works just like the original version does.
Would be great if you could spare an evening to check if anything looks suspicious.

natario1:
Will do this weekend, thanks!

natario1:
- `ZoomApi.kt` : do we need `JvmStatic` / `JvmField` on the companion object fields?
- `ZoomApi.kt` : annotated fields like `@get:Zoom` should also have the setter annotated, so I think you can simply use `@Zoom`?
- `ZoomEngine.kt` L 339: Wrong method name in replaceWith
- `ZoomEngine.kt` L 797: Should be `newZoom`, not `zoom`

Some Kotlinish nits,

- Engine: can mListeners be a mutableListOf instead of ArrayList?
- Engine: can you use forEach { } when iterating (dispatchOnIdle & dispatchOnMatrix)
- Logger: the `string` function could be one line with kotlin's `joinToString()` with a space separator, solves the trim problem also

Could you fix/check these? Should be easy. The rest seems cool (at a first glance, honestly I can't process 2k lines with my eyes on two browser tabs)

natario1:
Last thing which is important, sorry. 

For most files git took this change as a delete + create new file. This is wrong because it means that all git history is lost for these files. Can you fix this?

I would do this
- fix stuff from my comment above
- push to GitHub
- take a new clean branch from master
- rename file extensions from `java` to `kt`
- make a commit. Git should understand this is a rename
- Now edit these files, remove java content and paste kotlin content from GitHub
- Also paste changes to gradle files (or pick commits)
- Force push this new branch into feature/kotlin

Gave you access to the repo now

markusressel:
> * `ZoomApi.kt` : do we need `JvmStatic` / `JvmField` on the companion object fields?

Not on `const` fields.

> * `ZoomApi.kt` : annotated fields like `@get:Zoom` should also have the setter annotated, so I think you can simply use `@Zoom`?

Seems plausible.
> * `ZoomEngine.kt` L 339: Wrong method name in replaceWith

Right, missed that. I also removed the shorthand and replaced it with a default parameter and `@JvmOverloads`.
> * `ZoomEngine.kt` L 797: Should be `newZoom`, not `zoom`

I don't see any zoom there, am I missing something?
> 
> Some Kotlinish nits,
> 
> * Engine: can mListeners be a mutableListOf instead of ArrayList?

Sure.
> * Engine: can you use forEach { } when iterating (dispatchOnIdle & dispatchOnMatrix)

Yep. I removed the local variable `val matrix = matrix` in `dispatchOnMatrix` too, or is this of any importance to use a local variable there?
> * Logger: the `string` function could be one line with kotlin's `joinToString()` with a space separator, solves the trim problem also

Well the empty string special case is an additional line but I replaced the other stuff with `joinToString()`.
> 
> Could you fix/check these? Should be easy. The rest seems cool (at a first glance, honestly I can't process 2k lines with my eyes on two browser tabs)

No problem, we can always enhance it later :)

natario1:
Sorry I was looking at lines in this commit: https://github.com/natario1/ZoomLayout/pull/38/commits/499f9aca8469c2eaf29f76080df4e3142f5ac103 . The if (zoom > 0) condition should be on newZoom

markusressel:
> Sorry I was looking at lines in this commit: [499f9ac](https://github.com/natario1/ZoomLayout/commit/499f9aca8469c2eaf29f76080df4e3142f5ac103) . The if (zoom > 0) condition should be on newZoom

Ah yes I fixed that already.
I don't have time this evening anymore but I will get around the git commit thingy asap in the coming days (christmas is kind of busy).

markusressel:
I renamed the files using `git mv` because git was not able to pick up file renaming automatically.
When comparing the commits on GitHub files are shown as edited, but on the "Files Changed" Tab they are (partly) still shown as deleted... I think github is broken here, the commits should be fine.

Anything else to consider?

natario1:
Yes, looks like it works. It's fine, will merge now! Thank you for the effort!

natario1 (approved):

natario1:
As notes for the future, to release a new version, you just bump version in build.gradle and README, merge, then create a GitHub release, add who did what + PR links in the release notes, and Travis should publish to Bintray. Version names should be vX.X.X or Travis ignores them.

I tend to squash commits in PRs when merging. I just used the rebase option now because there was the rename thing and we would probably have lost history again.

PR #49, opened 2018-08-28
Update AndroidManifest.xml

bkhall:
The main objective here is to remove the allowBackup and supportsRtl tags; they were causing manifest merge errors.

But the most correct fix is to remove the application tag completely. Libraries don't need it.


natario1 (approved):

natario1:
Thanks !

Issue #32, opened 2018-06-12
Inconsistent layout for different Android device density (or OS version)


zahid-arbisoft:
I have a image view in child and for 420dpi device it is showing perfect centered vertical in the zoomlayout.

Same code on 560dpi device moves image to top (almost 80dp) behind a view on top of the screen. Image size is 1360 x 910. 

Can you give me some hint, i want to fix it myself.

Devices used:

**Update**
This issue could be because of OS version. Nexus 6P emulator (API 26) has no issue. Nexus 6P emulator (API 23) has the issue.

```
<LinearLayout
    xmlns:android="http://schemas.android.com/apk/res/android"
    xmlns:app="http://schemas.android.com/apk/res-auto"
    xmlns:tools="http://schemas.android.com/tools"
    android:layout_width="match_parent"
    android:layout_height="match_parent"
    android:orientation="vertical"
    android:weightSum="10">

    <com.otaliastudios.zoom.ZoomLayout
        android:layout_width="match_parent"
        android:layout_height="0dp"
        android:layout_weight="6"
        android:background="@color/colorAccent"
        app:maxZoom="3.0"
        app:minZoom="1.0">

        <LinearLayout
            android:layout_width="match_parent"
            android:layout_height="wrap_content">

            <ImageView
                android:layout_width="match_parent"
                android:layout_height="wrap_content"
                android:src="@drawable/img"/>
        </LinearLayout>

    </com.example.syed.draganddrop.engine.ZoomLayout>

    <LinearLayout
        android:layout_width="match_parent"
        android:layout_height="0dp"
        android:layout_weight="4"
        android:background="@android:color/darker_gray">

        <TextView
            android:layout_width="match_parent"
            android:layout_height="match_parent"
            android:gravity="center"
            android:textColor="@android:color/white"/>
    </LinearLayout>

</LinearLayout>
```

natario1:
I think this was fixed by #62

Issue #22, opened 2018-04-16
Enclosing View is not properly rendered on Android 6 and lower


W3D3:
I am using a custom View with a Canvas that is made zoomable with help of this library and this works fine on Android 7 and higher. 

Here are 2 reference images with the ZoomLayout unzoomed and not panned.

<img width="435" alt="screen shot 2018-04-16 at 19 50 19" src="https://user-images.githubusercontent.com/1103924/38826233-7aaed30a-41af-11e8-8ded-054899be43f8.png">
The Version running on Android API 27 is behaving as expected and draws the Canvas in the size of the screen. The view is zoomable and the coordinate transformation works so that the clicks on different Hexagons get detected.

________________
<img width="407" alt="screen shot 2018-04-16 at 19 50 27" src="https://user-images.githubusercontent.com/1103924/38826231-7a8f332e-41af-11e8-8d65-7df049ec02e1.png">

This is the same View on Android API 23 or lower (shown is API Level 18) where the whole canvas seemed to have moved towards the top. When panning down there is a white part of the ZoomLayout area that should not be there and indicates that the canvas got moved towards the top becomes visible. Clicks on the canvas are also not getting the right x and y values and therefore don't hit any of my defined regions which makes the View useless with zoom.
This does not happen when I just put the View inside a LinearLayout or something similar.

![screenshot_1523901541](https://user-images.githubusercontent.com/1103924/38826608-a0e8c5de-41b0-11e8-8d52-7e7fc81d5329.png)

Here is the XML Code for the view:

```xml
<?xml version="1.0" encoding="utf-8"?>

<LinearLayout
    android:id="@+id/gridContainer"
    android:layout_height="match_parent"
    android:layout_width="match_parent"
    xmlns:android="http://schemas.android.com/apk/res/android"
    android:orientation="vertical">


    <Button xmlns:android="http://schemas.android.com/apk/res/android"
        android:id="@+id/button"
        android:layout_width="wrap_content"
        android:layout_height="wrap_content"
        android:text="Button" />

    <com.otaliastudios.zoom.ZoomLayout xmlns:android="http://schemas.android.com/apk/res/android"
        xmlns:app="http://schemas.android.com/apk/res-auto"
        android:id="@+id/zoomContainer"
        android:layout_width="match_parent"
        android:layout_height="match_parent"
        app:hasClickableChildren="false"
        app:maxZoom="3"
        app:maxZoomType="zoom"
        app:minZoom="1"
[...]

PR #62, opened 2018-12-11
Fixes view offset on Android 6 and lower

DennisBlock:
This fixes the view offset on Android 6 and lower. If the ZoomLayout has no clickable children the child is moved to the top of the screen . It seems like the views matrix is already translated in the lower versions and overriding it with a different one removes the inital translation. (Altough the views matrix says it's an identiy matrix). This behavior can be seen in the issues #32 and #22 or simply by setting the `hasClickableChildren` to false in the sample.


natario1:
Sorry for the delay. I think this makes sense, thank you!

natario1 (approved):

PR #65, opened 2018-12-25
ZoomLayout’s OnGlobalLayoutListener holds WeakReferences to its dependencies

### LINKED ISSUE #58: Unregister ZoomLayout's OnGlobalLayoutListener
### Environment

**Library version:** `com.otaliastudios:zoomlayout:1.3.0`
**Android version:** Android 9 (API 28), tested on a Google Pixel 2

### Issue Description

While testing an application that uses `ZoomLayout`, a memory leak was reported by [LeakCanary](https://github.com/square/leakcanary). It seems that the leak is caused by the `OnGlobalLayoutListener` registered in `ZoomLayout.addView()`, which is never unregistered.

### Leak Information

The leak info exported from LeakCanary can be found at square/leakcanary#1126. I had opened that issue because I thought that the leak was caused by the Android SDK and should therefore be added to LeakCanary's exclusion list.

### Tests with ZoomLayout

I then re-tested my application with a fork of ZoomLayout in which I modified the `OnGlobalLayoutListener` to unregister itself the first time it is invoked. No leaks were reported by LeakCanary, and everything in ZoomLayout seemed to be working as expected. Do you think that there could be any negative side effects with that modification?

dmazzoni:
Fixes #58


library/src/main/java/com/otaliastudios/zoom/ZoomLayout.kt
```diff
@@ -80,6 +80,24 @@ private constructor(context: Context, attrs: AttributeSet?, @AttrRes defStyleAtt
...
+        private val mEngine = WeakReference(engine)
+        private val mChildView = WeakReference(childView)
+
+        override fun onGlobalLayout() {
+
+            val engine = mEngine.get()
+            val childView = mChildView.get()
+
+            if (childView == null || engine == null) {
+                return
```
markusressel, line 100: When the `childView` is not `null` the listener should be able to remove itself, right?
We should definitely try to avoid keeping a dead listener around - I thought this was the initial issue? Or is it sufficient that the view references are `WeakReferences` and the Listener will be removed by the system and those null checks are only to prevent some of the "last" events to cause null exceptions?

dmazzoni: My first proposal in the discussion of #58 was indeed to have the listener remove itself the first time it is fired (provided that `childView` is not `null`), but @natario1 pointed out that such behavior would not be correct. I then implemented the `WeakReference` approach so as to avoid leaking the view in case the listener is not removed by the system.

natario1: Just curious, would this leak happen if, instead of static listener + WeakReferences, we just try to remove the listener during `ZoomLayout.onDetachedFromWindow()`? That would be cleaner and we don't leave a dead listener around.

markusressel: @dmazzoni I am not talking about removing it the first time but when the engine is null and the view is not. But a simple `return` doesn't remove the listener ever - except when the view is destroyed I guess. Still seems like a workaround, what about @natario1 ?

markusressel: @dmazzoni I would relly like to merge this PR but the current solution does not seem optimal and I don't know how to reproduce this issue so it would be pretty cool if you could test if detaching it in `ZoomLayout.onDetachedFromWindow()` as @natario1 suggested also fixes the issue. If not I would merge the current solution (if @natario1 approves) to get this resolved. We can always improve later.
Thx for your work and interest!

dmazzoni: @markusressel If it's ok I'll be able to have a look at it next week

dmazzoni:
Sorry for the late reply! I'm currently testing the implementation that removes the listener when  ```onDetachedFromWindow()``` is called.

At the time ```onDetachedFromWindow()``` is invoked the ```ViewTreeObserver``` originally used to add the listener is no longer alive: the child view's getter returns a different ```ViewTreeObserver``` instance, and keeping the old instance is useless because calling any method on it other than ```isAlive()``` yields an ```IllegalStateException```.

From this perspective this leak looks like a platform bug, as I would expect listeners to be automatically unregistered by the system upon replacing the ```ViewTreeObserver```, as we cannot do it anymore once it is no longer alive.

markusressel:
How is it possible though that it returns a different viewtreeobserver? And does the proposed change fix this then anyway? A simple return wouldn't detach the listener either. Would it be possible to keep a reference of the initial viewtreeobserver in the listener and check if the one returned by the child view has changed and then unregister itself? seems like a big ass workaround though...

natario1:
> At the time `onDetachedFromWindow()` is invoked the `ViewTreeObserver` originally used to add the listener is no longer alive: the child view's getter returns a different `ViewTreeObserver` instance

Have you tried to call remove on this new instance though? This should work looking at the View source code.

Thanks for spending time on this, it would be cool if you have a demo app where this can be reproduced.

dmazzoni:
The Android documentation states that the ViewTreeObserver is not guaranteed to be valid for the whole lifetime of the View. In this case, the behavior I observed is that the View's getter returns the reference to the new ViewTreeObserver (at least on API 28).

At this point it is too late to do anything because the new ViewTreeObserver cannot be used to unregister the old listener, but neither the old one can because it is no longer alive.

The proposed change doesn't fix this, it just avoids leaking the whole ZoomLayout together with the listener.

natario1:
> At this point it is too late to do anything because the new ViewTreeObserver cannot be used to unregister the old listener

Have you tried?

The View class seems to merge the listeners when it swaps observers.
My point is that on the first call, you get the dummy observer (See `View.mFloatingTreeObserver`). When hierarchy is attached this is replaced by the real one, but they are merged (See `View.dispatchAttachedToWindow()`).

When we are being detached, you should get the "real" one which still has us as listeners, so remove should work. This is because `onDetachedFromWindow` is called before clearing the real observer (See `View.dispatchDetachedFromWindow()`).

dmazzoni:
@natario1 I did try, but to no avail. i don't know either why the dead ViewTreeObserver is still referenced by the DecorView's attachInfo, but that may depend on our application. Anyway tomorrow I'll work on a sample project

natario1:
With `child.viewTreeObserver` or with `zoomLayout.viewTreeObserver` ? I think the latter should work. Sorry to bother - I'm just looking at source code trying to understand what's happening here.

dmazzoni:
No problem! Yesterday I only tried the former. Today I tried removing the listener via ```zoomLayout.viewTreeObserver``` and it seems to work fine, as LeakCanary is no longer reporting any leak.

If it's OK with you I'd close this PR and open a separate one with the implementation of listener removal in ```onDetachedFromWindow()```.

natario1:
Man, these things are tricky.

I think what we could do is remove the current logic and do as follows

- have ZoomLayout implement OnGlobalLayoutListener
- viewTreeObserver.addListener(this) in onAttachedToWindow
- viewTreeObserver.removeListener(this) on onDetachedFromWindow
- during on global layout, check that we have child and if we do, call that engine method

I think this is cleaner? For example it covers the case where someone could call zoomLayout.removeView(child). What do you think?

dmazzoni:
I also think it is cleaner, and have just successfully tested it (no leaks reported by LeakCanary).

PR #68, opened 2018-12-28
Feature/#66 pan while zoom

### LINKED ISSUE #66: Allow pan while pinching
# Current Behaviour
The focus point when pinching is fixed at the center of both touch points when the pinching gesture is started. This means the content can not be panned during the pinch gesture.

# Expected Behaviour
The focus point moves with the center of both touch points. 
When starting the pinch gesture the center is remembered and when moving both touch points in the same direction and speed (without changing their distance) the content should be panned without any zoom changes.

This allows to zoom more conveniently to a specific position instead of having to zoom to the desired level and adjusting view position afterwards using pan gesture.

markusressel:
fixes #66 

Wow so this took quite a while for me to get it to work but I'm quite happy with the results.
The hardest part was animating back to a non-overscrolled and non-overpinched state at the same time.
I use a "simulated" zoom state to calculate the pan fixes needed to be applied to get back into a valid state when the zoom is also changing. I will add a review comment to the mentioned line, please let me know if you know a better way to do this.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -23,6 +23,6 @@ android {
 }
 
 dependencies {
-    implementation "androidx.appcompat:appcompat:$supportLibVersion"
+    implementation "androidx.appcompat:appcompat:1.0.2"
     implementation project(':library')
 }
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,6 +10,8 @@ android {
         targetSdkVersion rootProject.ext.targetSdkVersion
         versionCode 1
         versionName "1.0"
+
+        setProperty("archivesBaseName", "ZoomLayout_v${versionName}_(${versionCode})")
     }
 
     buildTypes {
--- a/library/src/main/java/com/otaliastudios/zoom/AbsolutePoint.kt
+++ b/library/src/main/java/com/otaliastudios/zoom/AbsolutePoint.kt
@@ -4,8 +4,8 @@ package com.otaliastudios.zoom
  * This class represents an absolute point on the ZoomEngine canvas (or beyond it's bounds)
  */
 data class AbsolutePoint(
-        @ZoomApi.AbsolutePan var x: Float,
-        @ZoomApi.AbsolutePan var y: Float) {
+        @ZoomApi.AbsolutePan var x: Float = 0F,
+        @ZoomApi.AbsolutePan var y: Float = 0F) {
 
     /**
      * Set new coordinates
@@ -35,6 +35,15 @@ data class AbsolutePoint(
      */
     operator fun minus(absolutePoint: AbsolutePoint): AbsolutePoint {
         return AbsolutePoint(this.x - absolutePoint.x, this.y - absolutePoint.y)
+    }
+
+    /**
+     * Negate a point
+     *
+     * @return the negative value of this point
+     */
+    operator fun unaryMinus(): AbsolutePoint {
+        return AbsolutePoint(-this.x, -this.y)
     }
 
     /**
--- a/library/src/main/java/com/otaliastudios/zoom/AbsolutePoint.kt
+++ b/library/src/main/java/com/otaliastudios/zoom/AbsolutePoint.kt
@@ -14,9 +14,9 @@ data class AbsolutePoint(
      * @param y y-axis value
      */
     @JvmOverloads
-    fun set(@ZoomApi.AbsolutePan x: Float = this.x, @ZoomApi.AbsolutePan y: Float = this.y) {
-        this.x = x
-        this.y = y
+    fun set(@ZoomApi.AbsolutePan x: Number = this.x, @ZoomApi.AbsolutePan y: Number = this.y) {
+        this.x = x.toFloat()
+        this.y = y.toFloat()
     }
 
     /**
@@ -53,6 +53,16 @@ data class AbsolutePoint(
      */
     operator fun plus(absolutePoint: AbsolutePoint): AbsolutePoint {
         return AbsolutePoint(this.x + absolutePoint.x, this.y + absolutePoint.y)
+    }
+
+    /**
+     * Multiply every value in the point by a given factor
+     *
+     * @param factor the factor to use
+     * @return the multiplied point
+     */
+    operator fun times(factor: Number): AbsolutePoint {
+        return AbsolutePoint(factor.toFloat() * this.x, factor.toFloat() * this.y)
     }
 
 }
--- a/library/src/main/java/com/otaliastudios/zoom/AbsolutePoint.kt
+++ b/library/src/main/java/com/otaliastudios/zoom/AbsolutePoint.kt
@@ -1,0 +1,49 @@
+package com.otaliastudios.zoom
+
+/**
+ * This class represents an absolute point on the ZoomEngine canvas (or beyond it's bounds)
+ */
+data class AbsolutePoint(
+        @ZoomApi.AbsolutePan var x: Float,
+        @ZoomApi.AbsolutePan var y: Float) {
+
+    /**
+     * Set new coordinates
+     *
+     * @param x x-axis value
+     * @param y y-axis value
+     */
+    @JvmOverloads
+    fun set(@ZoomApi.AbsolutePan x: Float = this.x, @ZoomApi.AbsolutePan y: Float = this.y) {
+        this.x = x
+        this.y = y
+    }
+
+    /**
+     * Set new coordinates
+     *
+     * @param p the [AbsolutePoint] to copy values from
+     */
+    fun set(p: AbsolutePoint) {
+        set(p.x, p.y)
+    }
+
+    /**
+     * Substract a point from another point
+     *
+     * @param absolutePoint the point to substract
+     */
+    operator fun minus(absolutePoint: AbsolutePoint): AbsolutePoint {
+        return AbsolutePoint(this.x - absolutePoint.x, this.y - absolutePoint.y)
+    }
+
+    /**
+     * Add a point to another point
+     *
+     * @param absolutePoint the point to add
+     */
+    operator fun plus(absolutePoint: AbsolutePoint): AbsolutePoint {
+        return AbsolutePoint(this.x + absolutePoint.x, this.y + absolutePoint.y)
+    }
+
+}
--- a/library/src/main/java/com/otaliastudios/zoom/AbsolutePoint.kt
+++ b/library/src/main/java/com/otaliastudios/zoom/AbsolutePoint.kt
@@ -8,6 +8,13 @@ data class AbsolutePoint(
         @ZoomApi.AbsolutePan var y: Float = 0F) {
 
     /**
+     * Copy constructor
+     *
+     * @param point point to duplicate
+     */
+    constructor(point: AbsolutePoint) : this(point.x, point.y)
+
+    /**
      * Set new coordinates
      *
      * @param x x-axis value
--- a/library/src/main/java/com/otaliastudios/zoom/ZoomApi.kt
+++ b/library/src/main/java/com/otaliastudios/zoom/ZoomApi.kt
@@ -37,6 +37,13 @@ interface ZoomApi {
     @RealZoom
     val realZoom: Float
 
+
+    /**
+     * The current pan as an [AbsolutePoint].
+     * This field will be updated according to current pan when accessed.
+     */
+    val pan: AbsolutePoint
+
     /**
      * Returns the current horizontal pan value, in content coordinates
      * (that is, as if there was no zoom at all).
--- a/library/src/main/java/com/otaliastudios/zoom/ZoomApi.kt
+++ b/library/src/main/java/com/otaliastudios/zoom/ZoomApi.kt
@@ -254,7 +254,7 @@ interface ZoomApi {
      * @param realZoom the new real zoom value
      * @param animate  whether to animate the transition
      */
-    fun realZoomTo(realZoom: Float, animate: Boolean)
+    fun realZoomTo(@RealZoom realZoom: Float, animate: Boolean)
 
     /**
      * Which is the max zoom that should be allowed.
--- a/library/src/main/java/com/otaliastudios/zoom/ZoomApi.kt
+++ b/library/src/main/java/com/otaliastudios/zoom/ZoomApi.kt
@@ -60,6 +60,7 @@ interface ZoomApi {
      *
      * @see realZoom
      */
+    @Target(AnnotationTarget.FIELD, AnnotationTarget.FUNCTION, AnnotationTarget.LOCAL_VARIABLE, AnnotationTarget.PROPERTY, AnnotationTarget.VALUE_PARAMETER)
     @Retention(AnnotationRetention.SOURCE)
     annotation class RealZoom
 
@@ -68,6 +69,7 @@ interface ZoomApi {
      *
      * @see zoom
      */
+    @Target(AnnotationTarget.FIELD, AnnotationTarget.FUNCTION, AnnotationTarget.LOCAL_VARIABLE, AnnotationTarget.PROPERTY, AnnotationTarget.VALUE_PARAMETER)
     @Retention(AnnotationRetention.SOURCE)
     annotation class Zoom
 
@@ -78,6 +80,7 @@ interface ZoomApi {
      * @see panY
      * @see ScaledPan
      */
+    @Target(AnnotationTarget.FIELD, AnnotationTarget.FUNCTION, AnnotationTarget.LOCAL_VARIABLE, AnnotationTarget.PROPERTY, AnnotationTarget.VALUE_PARAMETER)
     @Retention(AnnotationRetention.SOURCE)
     annotation class AbsolutePan
 
@@ -88,6 +91,7 @@ interface ZoomApi {
      * @see panY
      * @see AbsolutePan
      */
+    @Target(AnnotationTarget.FIELD, AnnotationTarget.FUNCTION, AnnotationTarget.LOCAL_VARIABLE, AnnotationTarget.PROPERTY, AnnotationTarget.VALUE_PARAMETER)
     @Retention(AnnotationRetention.SOURCE)
     annotation class ScaledPan
 
@@ -97,6 +101,7 @@ interface ZoomApi {
      * @see zoom
      * @see realZoom
      */
+    @Target(AnnotationTarget.FIELD, AnnotationTarget.FUNCTION, AnnotationTarget.LOCAL_VARIABLE, AnnotationTarget.PROPERTY, AnnotationTarget.VALUE_PARAMETER)
     @Retention(AnnotationRetention.SOURCE)
     @IntDef(TYPE_ZOOM, TYPE_REAL_ZOOM)
     annotation class ZoomType
--- a/library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
+++ b/library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
@@ -497,7 +497,7 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
             LOG.i("onSizeChanged: newBaseZoom:", mBaseZoom, "newZoom:", zoom)
             @Zoom val newZoom = checkZoomBounds(zoom, false)
             LOG.i("onSizeChanged: scaleBounds:", "we need a zoom correction of", newZoom - zoom)
-            if (newZoom != zoom) applyZoom(newZoom, false)
+            if (newZoom != zoom) applyZoom(newZoom, allowOverPinch = false)
 
             // pan based on transformation gravity.
             @ScaledPan val newPan = computeBasePan()
@@ -505,7 +505,7 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
             @ScaledPan val deltaY = newPan[1] - scaledPanY
             if (deltaX != 0f || deltaY != 0f) applyScaledPan(deltaX, deltaY, false)
 
-            ensurePanBounds(false)
+            ensurePanBounds(allowOverScroll = false)
             dispatchOnMatrix()
             if (!mInitialized) {
                 mInitialized = true
@@ -530,12 +530,12 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
             // This is a possible matrix update.
             @Zoom val newZoom = checkZoomBounds(zoom, false)
             LOG.i("onSizeChanged: scaleBounds:", "we need a zoom correction of", newZoom - zoom)
-            if (newZoom != zoom) applyZoom(newZoom, false)
+            if (newZoom != zoom) applyZoom(newZoom, allowOverPinch = false)
 
             // If there was any, pan should be kept. I think there's nothing to do here:
             // If the matrix is kept, and real zoom is kept, then also the real pan is kept.
             // I am not 100% sure of this though, so I prefer to call a useless dispatch.
-            ensurePanBounds(false)
+            ensurePanBounds(allowOverScroll = false)
             dispatchOnMatrix()
         }
     }
@@ -808,10 +808,10 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
 
                 applyZoomAndAbsolutePan(newZoom,
                         panX + mCurrentAbsFocusOffset.x, panY + mCurrentAbsFocusOffset.y,
-                        zoomTargetX = detector.focusX,
-                        zoomTargetY = detector.focusY,
                         allowOverScroll = true,
-                        allowOverPinch = true)
+                        allowOverPinch = true,
+                        zoomTargetX = detector.focusX,
+                        zoomTargetY = detector.focusY)
                 return true
             }
             return false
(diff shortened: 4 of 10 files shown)


library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -753,54 +764,154 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
...
                 val factor = detector.scaleFactor
                 val newZoom = zoom * factor
-                applyPinch(newZoom, mAbsTargetX, mAbsTargetY, true)
+
+                // apply the new zoom and pan values accordingly
+                applyZoomAndAbsolutePan(newZoom,
+                        mCurrentAbsTargetX, mCurrentAbsTargetY,
+                        allowOverScroll = true,
+                        allowOverPinch = true)
                 return true
```
markusressel, line 851: This is the stuff (above code) that allows pan and zoom to be applied simultaniously when using pinch gesture.

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -753,54 +764,154 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
...
+                    // select zoom pivot point based on what edge of the screen is currently overscrolled
+                    var zoomTargetX = calculateZoomPivotPoint(true, fixPanX)
+                    var zoomTargetY = calculateZoomPivotPoint(false, fixPanY)
+
+                    // calculate the new pan position
+                    var newPanX = panX + fixPanX
+                    var newPanY = panY + fixPanY
+
+                    if (newZoom.compareTo(zoom) != 0) {
+                        // we have overpinched. to calculate how much pan needs to be applied
```
markusressel, line 894: Here is the zoom "simulation" part. The problem I faced was that I couldn't just calculate pan fix with the current zoom and apply it at the same time as the new zoom level as the pan fix calculated by the `checkPanBounds` method is an absolute pan value that is always located at the top left of the view (I think?). Therefore this absolute position depends on the current zoom level - so the calculated fix is not applicable to the state after the overpinched zoom has been changed.

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -753,54 +764,154 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
...
+                        // recalculate pivot point based on the recalculated pan fix
+                        zoomTargetX = calculateZoomPivotPoint(true, fixPanX)
+                        zoomTargetY = calculateZoomPivotPoint(false, fixPanY)
+
+                        // recalculate new pan location using the simulated target zoom level
+                        newPanX = panX + fixPanX
+                        newPanY = panY + fixPanY
+
+                        // revert simulation
+                        applyZoom(oldZoom, true, true, zoomTargetX, zoomTargetY)
```
markusressel: Those `applyZoom` calls currently trigger listeners. I don't know if that's a problem but I think it should be avoided as it's just a simulation to calculate pan fix numbers?

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -1240,29 +1364,30 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
...
+     * @param zoom        the new zoom
      * @param targetX        the target X in abs value
      * @param targetY        the target Y in abs value
      * @param allowOverPinch whether to overPinch
      */
-    private fun applyPinch(@Zoom newZoom: Float, @AbsolutePan targetX: Float, @AbsolutePan targetY: Float,
+    private fun applyPinch(@Zoom zoom: Float, @AbsolutePan targetX: Float, @AbsolutePan targetY: Float,
                            allowOverPinch: Boolean) {
-        var newZoom = newZoom
+        // TODO: this method is currently unused, should it be removed? The doc mentions buggy translations :S
```
markusressel: You used this method to apply zoom **only** in the `PinchListener`.
I don't know if it is of any other value and should be kept as some kind of helper?

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -1044,25 +1155,23 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
...
-    private fun animateZoom(@Zoom newZoom: Float, allowOverPinch: Boolean) {
-        var newZoom = newZoom
-        newZoom = checkZoomBounds(newZoom, allowOverPinch)
+    private fun animateZoom(@Zoom zoom: Float, allowOverPinch: Boolean) {
         if (setState(ANIMATING)) {
             mClearAnimation = false
             val startTime = System.currentTimeMillis()
-            @Zoom val startZoom = zoom
-            @Zoom val endZoom = newZoom
+            @Zoom val startZoom = this.zoom
```
markusressel, line 1267: I'm sorry - there are a number of small refactorings and variable renamings in this PR.
Android Studio was getting on my nerves with duplicate variable names (parameter and local variable with the same name) and I had a hard time understanding what the `time` variable actually is used for so I changed those things.
If there is anything that speaks against this I can try to reverse those changes - but if you want this please let me know exactly what I need to reverse.

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -753,54 +764,154 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
...
                 }
 
                 // Having both overPinch and overScroll is hard to manage, there are lots of bugs if we do.
                 val factor = detector.scaleFactor
                 val newZoom = zoom * factor
-                applyPinch(newZoom, mAbsTargetX, mAbsTargetY, true)
+
+                // apply the new zoom and pan values accordingly
+                applyZoomAndAbsolutePan(newZoom,
+                        mCurrentAbsTargetX, mCurrentAbsTargetY,
```
markusressel: I just realized this pinch is missing the zoom target

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -753,54 +764,154 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
...
+                scaledFocusX = -detector.focusX
+                scaledFocusY = -detector.focusY
+                LOG.i("onScale:", "Setting focus.", "detectorFocusX:", scaledFocusX, "detectorFocusX:", scaledFocusY)
+
+                // Account for current pan.
+                scaledFocusX += scaledPanX
+                scaledFocusY += scaledPanY
+
+                mCurrentAbsTargetX = panX
+                mCurrentAbsTargetY = panY
```
markusressel: ~This should be nInitialAbsTargetX if I'm not mistaking. this might cause the little jump at the beginning when pinching fast.~

markusressel: Forget this, I was mistaking.
I though about this pan shifting while pinching again and it's now fixed.
Its much more smooth, no initial jump and works in all situations I could come up with.

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
markusressel: fixed

library/src/main/java/com/otaliastudios/zoom/AbsolutePoint.kt
```diff
@@ -0,0 +1,49 @@
+package com.otaliastudios.zoom
+
+/**
+ * This class represents an absolute point on the ZoomEngine canvas (or beyond it's bounds)
+ */
+data class AbsolutePoint(
```
natario1, line 6: These make everything cleaner, thanks.
Are they exposed? If not let's make them internal.

library/src/test/java/com/otaliastudios/zoom/PointTests.kt
```diff
@@ -0,0 +1,43 @@
+package com.otaliastudios.zoom
+
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.TestInstance
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
```
natario1: Thanks for tests!

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -1235,34 +1356,81 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
         dispatchOnMatrix()
     }
 
+    /**
+     * Converts an [AbsolutePoint] to a [ScaledPoint]
+     */
+    private fun AbsolutePoint.toScaled(): ScaledPoint {
```
natario1, line 1493: Nice. I think we could finally remove `resolvePan` and `unresolvePan` and create extensions of the `Float` class with the same names?

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
natario1: I think it's fine

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
natario1: I should take a better look to understand what's happening. Anyway I just tested the app and it's not super smooth so there must be a bug somewhere. 

For instance if I pinch out (move fingers closer) then release, it suddenly moves to top-left then goes back. It seems to do this during pinch+pan also. Do you have any idea?

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
natario1: I mean to call them toScaled() and toAbsolute() of course.

library/src/test/java/com/otaliastudios/zoom/PointTests.kt
markusressel: Well they are pretty basic so far, I just wanted to get the basic structure working. 

Initially I wanted to test methods of the engine but I failed to create a test that has a valid context that is needed by the engine :-(
I think this can be another PR though.

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
markusressel: You are right when zooming out more than the size of the content (so there are gray areas all around the image) the view does jump when pinching out.
I will have a look at that.

I don't know of any other jumping though if this is not the scenario you are describing.

library/src/main/java/com/otaliastudios/zoom/AbsolutePoint.kt
markusressel: As `panX` and `panY` are exposed I also exposed `pan` which is an `AbsolutePoint`. I can. remove that if it should only be an internal representation but I think the easier handling of points can also be beneficial to devs using this library - don't you think?

natario1: Yes, cool!

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
markusressel: Sure.

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
markusressel: I think I fixed it using the center of `mContentRect` as the pivot point when zooming in from underpinch. @natario1 Could you have another look?

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
markusressel: I added a boolean parameter `triggerListeners` to the relevant methods to prevent triggering them when changes are only simulated.

app/build.gradle
```diff
@@ -10,6 +10,8 @@ android {
         targetSdkVersion rootProject.ext.targetSdkVersion
         versionCode 1
         versionName "1.0"
+
+        setProperty("archivesBaseName", "ZoomLayout_v${versionName}_(${versionCode})")
```
natario1, line 14: What will this do?

library/src/main/java/com/otaliastudios/zoom/ZoomApi.kt
```diff
@@ -88,6 +91,7 @@ interface ZoomApi {
      * @see panY
      * @see AbsolutePan
      */
+    @Target(AnnotationTarget.FIELD, AnnotationTarget.FUNCTION, AnnotationTarget.LOCAL_VARIABLE, AnnotationTarget.PROPERTY, AnnotationTarget.VALUE_PARAMETER)
     @Retention(AnnotationRetention.SOURCE)
     annotation class ScaledPan
```
natario1, line 103: I regret using these name, something like AbsoluteCoord would have been better. But we will change in another version

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -150,9 +150,14 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
...
     @RealZoom
     override val realZoom: Float
-        @RealZoom
         get() = zoom * mBaseZoom
 
+    /**
+     * The current pan as an [AbsolutePoint]
+     */
+    private val pan: AbsolutePoint
+        get() = AbsolutePoint(panX, panY)
```
natario1: Are these pan and scaledPan values accessed frequently?
If they are it might be worth to avoid all these new instances by caching the value (have a private _pan and use _pan.set() in this getter), since touch events are a lot.

By the way I thought from another comment that these two were public?

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -625,10 +637,25 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
...
     }
 
+    /**
+     * Uses [checkPanBounds] to calculate the pan correction that
+     * needs to be applied to get into a valid state.
+     *
+     * @return the scaled pan fix values for x- and y-axis
+     */
+    private fun getPanCorrection(): ScaledPoint {
+        return ScaledPoint(
```
natario1: See if it's better to avoid this new instance as well depending on how frequently this is accessed

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
natario1: Seems to work great now! Nice work!

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -773,58 +806,171 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
...
+     * @return the currently set lower boundary for minimum zoom value
+     */
+    @Zoom
+    private fun getMinZoom(): Float = mMinZoom.toZoom(mMinZoomMode)
 
     private inner class FlingScrollListener : GestureDetector.SimpleOnGestureListener() {
 
         override fun onDown(e: MotionEvent): Boolean {
+            // cancel any animation that might still be running
+            setState(NONE)
```
natario1: I see some issues with this: zoom in, then fling fast toward one end, then put your finger when the content is overscrolled. It does not always come back.

Might be easy to fix, but we can leave it for another PR I think.

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -1056,86 +1202,118 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
...
+        if (zoom <= getMinZoom()) {
+            zoomTo(getMinZoom(), animate = true)
         }
     }
 
     //endregion
 
     //region Apply values
 
+    private val mCancelAnimationListener = object : Animator.AnimatorListener {
```
natario1, line 1232: Do you think we should setState(NONE) onAnimationCancel also?

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -1056,86 +1202,118 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
...
+            @Zoom val endZoom = checkZoomBounds(zoom, allowOverPinch)
+            val zoomAnimator = ValueAnimator.ofFloat(startZoom, endZoom)
+            zoomAnimator.prepare()
+            zoomAnimator.addUpdateListener {
+                LOG.v("animateZoom:", "animationStep:", it.animatedFraction)
+                if (mClearAnimation) {
+                    it.cancel()
                 }
-            })
+                mContainer.postOnAnimation {
```
natario1: You don't need this post here, I think the animator already runs in an animation step.

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -1056,86 +1202,118 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
...
+                    PropertyValuesHolder.ofFloat(
+                            "zoom",
+                            startZoom, endZoom)
+            )
+            animator.prepare()
+            animator.addUpdateListener {
+                if (mClearAnimation) {
+                    it.cancel()
+                }
+                mContainer.postOnAnimation {
```
natario1: Same here, post is not needed. Nice approach

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -1147,60 +1325,63 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
      * @param deltaY          a scaled delta
      * @param allowOverScroll whether to overscroll
      */
-    private fun animateScaledPan(@ScaledPan deltaX: Float, @ScaledPan deltaY: Float,
+    private fun animateScaledPan(@AbsolutePan deltaX: Float, @ScaledPan deltaY: Float,
```
natario1: I think this was an error

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -1147,60 +1325,63 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
...
+                startValue + (endValue - startValue) * fraction - scaledPan
+            }, startPan, endPan)
+            panAnimator.prepare()
+            panAnimator.addUpdateListener {
+                LOG.v("animateScaledPan:", "animationStep:", it.animatedFraction)
+                if (mClearAnimation) {
+                    it.cancel()
                 }
-            })
+                mContainer.postOnAnimation {
```
natario1: Same as before, post not needed

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
natario1: If it's unused we can remove it.

markusressel: Ok

app/build.gradle
markusressel: When building an apk (release or debug) the resulting apk file will be called this way.
I have too much `app-debug.apk` files in my life already :stuck_out_tongue:

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
markusressel: I made them public now, don't know why they weren't.
Yes using the same Object is probably better for performance - I changed that.

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
markusressel: jup good catch!

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
markusressel: ok

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
markusressel: ok

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
markusressel: ok

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
markusressel: done

library/src/main/java/com/otaliastudios/zoom/ZoomApi.kt
markusressel: Those names still confuse me :grin: 
They are also kind of useless since the IDE seems to do nothing to check these.
Even when passing wrong values to a function with annotated parameters there is no warning.
That's another reason why I opted to create data objects for them (although you can still pass in wrong values on their constructors...)

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
markusressel: ok I will just revert this little change in this PR then.

library/src/main/java/com/otaliastudios/zoom/ZoomApi.kt
natario1: Ah, without them it would be worse I think... There should be an android annotation to tell the IDE that these two annotations exclude each other, but I don't know how.

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
markusressel: Well the animation is only cancelled when `mClearAnimation` is set to `true` which is done in `setState(NONE)` ... so currently it's the other way around - the state change causes the animationListener to trigger onCancel. It wouldn't hurt though I guess since it will already be in that state.

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
natario1: I didn't know you could do what you did :-) just one last change, can they be vals of ZoomApi overriden in the engine?

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
natario1: Well it's not important then, as you wish

library/src/main/java/com/otaliastudios/zoom/ZoomApi.kt
markusressel: Yes they do help indeed! But not to the extend I'd like them to :stuck_out_tongue:

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
markusressel: Well the `pan` val can, the `scaledPan` depends on `mTransformedRect` so I don't think so?

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -1151,56 +1329,57 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
...
+            val endPan = startPan + ScaledPoint(deltaX, deltaY)
+
+            val panAnimator = ValueAnimator.ofObject(TypeEvaluator { fraction, startValue: ScaledPoint, endValue: ScaledPoint ->
+                startValue + (endValue - startValue) * fraction - scaledPan
+            }, startPan, endPan)
+            panAnimator.prepare()
+            panAnimator.addUpdateListener {
+                LOG.v("animateScaledPan:", "animationStep:", it.animatedFraction)
+                if (mClearAnimation) {
+                    it.cancel()
```
natario1, line 1359: Sorry I forgot this, we need a return after these it.cancel() to be as it was.

markusressel: right, I changed it in all 3 places.

natario1 (approved):
This is good IMO! It was a great addition, thanks.

markusressel:
Since you already approved this PR and I only fixed a small dev-merge mistake I will merge this PR now.
If any merge conflicts come up in #74 because of this and you are unsure what to do just let me know.
Cheers! :tada:

PR #70, opened 2018-12-29
added API methods: setFlingEnabled, setAllowFlingInOverscroll

### LINKED ISSUE #69: Fling detection doesn't work when overscrolled
## Current Behaviour
1. scroll with a single finger to the lower right corner of the view
2. try to fling up with a slight tendency to the right border causing a slight overscroll on the right edge
3. release touch
 => the upwards motion stops immediately and an animation is started to return to a non-overscrolled state

## Expected Behaviour
A fling gesture is detected regardless of the slight overscroll. The fling is executed and (as is currently the case) also automatically returns to a non-overscrolled state but keeps the initial upwards motion.

markusressel:
fixes #69 

Thanks to @kevinadhitama for your initial work.

@natario1 I saw in the code that you had issues with some kind of artifacts when allowing fling events while overscrolled. I tested it on my device (running Android 9...) and couldn't cause any unwanted effects. To make sure the current behaviour is not changed but people who want to use it anyway (like me) can do that I added a flag and API method to change this behavior.

### DIFF
--- a/library/src/main/res/values/attrs.xml
+++ b/library/src/main/res/values/attrs.xml
@@ -1,27 +1,29 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <declare-styleable name="ZoomEngine">
-        <attr name="hasClickableChildren" format="boolean"/>
-        <attr name="overScrollHorizontal" format="boolean"/>
-        <attr name="overScrollVertical" format="boolean"/>
-        <attr name="overPinchable" format="boolean"/>
+        <attr name="hasClickableChildren" format="boolean" />
+        <attr name="overScrollHorizontal" format="boolean" />
+        <attr name="overScrollVertical" format="boolean" />
+        <attr name="overPinchable" format="boolean" />
         <attr name="zoomEnabled" format="boolean" />
+        <attr name="flingEnabled" format="boolean" />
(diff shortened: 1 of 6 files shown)


library/src/main/res/values/attrs.xml
```diff
@@ -1,27 +1,29 @@
...
-        <attr name="overScrollHorizontal" format="boolean"/>
-        <attr name="overScrollVertical" format="boolean"/>
-        <attr name="overPinchable" format="boolean"/>
+        <attr name="hasClickableChildren" format="boolean" />
+        <attr name="overScrollHorizontal" format="boolean" />
+        <attr name="overScrollVertical" format="boolean" />
+        <attr name="overPinchable" format="boolean" />
         <attr name="zoomEnabled" format="boolean" />
+        <attr name="flingEnabled" format="boolean" />
+        <attr name="allowFlingInOverscroll" format="boolean" />
```
markusressel, line 10: These two lines are the only addition **in this file**, other stuff is just reformatted.

natario1:
Cool! Honestly I can't remember what kind of issues. Was something I was seeing in some other app, not the demo. Let's use this way and we'll fix artifacts when they are reproducible 👍

natario1 (approved):

PR #71, opened 2019-01-02
Added the possibility to have a purely gravity based transformation

sgravel-dev:
I'm currently using your great ZoomLayout to have a 2D zoomable scroll view in Xamarin Forms for a customer.  However, because of your design decision to always center a view that is smaller than the viewport, I could not have feature parity with iOS's native zoom on scroll views and could not implement the desired layout.  I therefore added a new transformation mode called GRAVITY that will not scale and that respects the `TransformationGravity` passed in.

For instance, a `bottom|right` gravity with a GRAVITY transformation will always place the content at the bottom right and panning will only become available once the view is larger than the zoom layout.  This is essential for a report-type grid I'll be using this with.

I opted for this API to keep the NONE transformation mode as-is but it could be done using the GRAVITY mode and a `center` gravity flag.  You should consider merging both modes in the future.


markusressel:
Thanks for your PR!

My question would be: why is this not working with the NONE transformation currently - or is it?
In my mind it would be much more logical if this was the default behaviour. So in my opinion this should be possible without the introduction of a new Transformation mode or am I missing something?

**EDIT:**
Just to be clear my suggestion would be to use `zoomLayout.setTransformation(ZoomApi.TRANSFORMATION_NONE, Gravity.RIGHT | Gravity.CENTER);` or similar.

I would even go as far as to changing the constant names of the transformation constants to seperate it completely from the gravitation aspect but I guess @natario1 won't be happy about that as it would require breaking changes. Isn't the transformation supposed to be only "zoom" related?

sgravel-dev:
Hello,
The difference between the NONE transformation mode and the GRAVITY is the way it handles positioning of the content when it is smaller than the container.  The NONE mode always centers the content within the container whereas the new GRAVITY mode will honour the gravity flags, effectively changing the zooming pivot.  It is somewhat easier to try than to explain ;)

So, yes, if you ask me, the NONE mode is kind of a bastard child.  It does consider gravity but only if the content is larger than the container when initialized.  Apart from that, it does not impact the rendering transform.  With that said, it is called a transform not a zoom mode so I see no reason it would only affect the zoom.  The way I see it, the current NONE mode should be renamed CENTER and then we could introduce the new GRAVITY mode.  As I mentionned in the PR, I propose introducing the new mode to prevent API changes to the NONE mode but the project’s owner is free to decide if he prefers a breaking change.

natario1:
Thanks, will make some points:

The `transformation` semantics refer to what happens when the view size is initially set (or if you explicitly call setSize with `transformation = true`). Applying the transformation basically means let's go back to the starting point, more or less. It's a kind of keyframe.

So choosing a transformation should not change the engine behavior when scrolling, flinging or whatever else. The "center content when smaller than container" is a core behavior of the engine (and should be, in my opinion, OK for 99.9% of users) but it is not a "transformation" property.

So if you need to change this I would rather add a `disallowCenterWhenContentIsSmaller()` explicit flag, that just modifies `getPanCorrection()` to return the same values as you did. I would also leave the name negative as, again, IMO, there are not so many usages about it.

If you want to go future-proof & make it more elegant, the API could be like `setSmallerPolicy` (can't think of a better name right now, docs should suffice). SmallerPolicies would control what happens when content is smaller than container, and prob. should only act in the `getPanCorrection()` method. Then you could define:

- `SMALLER_POLICY_CENTER` // Current behavior, always center when smaller
- `SMALLER_POLICY_FROM_TRANSFORMATION` // Respect the transformation gravity

Does this make sense to you?

sgravel-dev:
Well to answer your question, no it doesn’t really make sense to me.  Which doesn’t mean I really care much about making changes to fit your proposal.  Nevertheless, here is why it doesn’t make sense to me.

1) The nomenclature isn’t clear.  What you have is not a tranformation mode but a scale with a base panning.  I’m not sure why one should set a “transform“ instead of using the `PanTo` method with a ScaleType enum instead.
2) Gravity is a different concept that you reuse here but with a different meaning.  Again, I’m unsure why you haven’t used the Android GravityFlags instead and giving it a new meaning creates confusion.  It should also be set on the content view using `layout_gravity` not on the container to be inline with other layouts.
3) I doubt centering the view really fits 99.9% of users as it is not the default behavior on iOS.  I’ll be using your library to allow scaling of a spreadsheet-like report.  This seems like a common use-case of a zoomable scrollview.  The Android ScrollView also does not center it’s content by default and I expected the same behavior.

With that said, I just need a feature and you will be the one maintaining it so I’ll go as you propose.

natario1:
The transformation nomenclature might not be the best, but what it is is a keyframe that the engine reach at certain points, for example at startup or when the container size is set.

It defines a resting position, not the content "transform" during operations. It is a useful feature in my opinion in conjunction with the minZoom.

The "center when smaller" policy operates on a different level, it is always triggered during scroll events, flings and pinches. And this is what you need to change, right?

So let's do it, it will be a good improvement, but it doesn't fit the "transformation" concept as it is defined now. 

We don't do XML fancyness because the engine is a abstract piece and the content is not necessarily a View. We also can't question the lib structure as it is being used and we can't change its surface/behavior.

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -615,23 +620,30 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
...
+                                 @ScaledPan basePanValue: Float): Float {
         @ScaledPan val tolerance = overScroll.toInt()
         var min: Float
         var max: Float
         if (contentSize <= viewSize) {
-            // If contentSize <= viewSize, we want to stay centered.
-            // Need a positive translation, that shows some background.
-            min = (viewSize - contentSize) / 2f
-            max = (viewSize - contentSize) / 2f
+            if(mSmallerPolicy == ZoomApi.SMALLER_POLICY_FROM_TRANSFORMATION) {
```
markusressel: Using a `when` statement for this kind of check is much more elegant imo.

sth. like:
```
// contentSize <= viewSize, use mSmallerPolicy to decide how to correct it
when (mSmallerPolicy) {
    ZoomApi.SMALLER_POLICY_FROM_TRANSFORMATION -> {
        min = basePanValue
        max = basePanValue
    }
    else -> {
        // we want to stay centered.
        // Need a positive translation, that shows some background.
        min = (viewSize - contentSize) / 2f
        max = (viewSize - contentSize) / 2f
    }
}
```

library/src/main/java/com/otaliastudios/zoom/ZoomApi.kt
```diff
@@ -310,5 +326,17 @@ interface ZoomApi {
...
+
+        /**
+         * Constant for [ZoomApi.setSmallerPolicy].
+         * Content will be centered when it is smaller than it's container
+         */
+        const val SMALLER_POLICY_CENTER = 0
+
+        /**
+         * Constant for [ZoomApi.setSmallerPolicy].
+         * Content will be respect the transformationGravity when it is smaller than it's container
```
markusressel: typo: will ~be~ respect

markusressel (approved):
Looks good to me, can this be merged @natario1 ?

natario1:
It works well IMO! @LRP-sgravel could you just add some notes in README about it?

Notes for the future,
- computeBasePan should not allocate an array each time it is called
- if anyone comes up with a better name than `SmallerPolicy`, please suggest before the next release :-)
- a third nice type would be SMALLER_POLICY_NONE, which lets the content freely move inside the container bounds, like when you move a sticker over an image to place it

On the same line, it would be super cool one day for the engine to accept multiple children, and do bounds detection, so the touch events are forwarded to that specific child (which in SMALLER_POLICY_NONE would work like a Instagram story layout, where you can put and deal with multiple stickers). Maybe something like a ZoomLayout holding multiple engines, not sure. Just random thoughts.

sgravel-dev:
Done

sgravel-dev:
Cool!
You might want to link to the Xamarin binding library I created (https://www.nuget.org/packages/lrp.xamarin.zoomlayout/) or the Xamarin Forms control I developed where I use your library (https://www.nuget.org/packages/LRP.Xamarin.Forms.ZoomScrollView/)

natario1:
Thank you @LRP-sgravel this was a cool addition. Note that before release we will rename `from_transformation` XML to `fromTransformation` for consistency.

Feel free to add a `Made with ZoomLayout` list at the end of README :-)

PR #74, opened 2019-01-07
Add SMALLER_POLICY_NONE

natario1:
Adds a new policy that lets the small content be moved around.


README.md
```diff
@@ -164,16 +166,22 @@ cropped along some dimension, the engine will also apply a translation according
...
+[Smaller Policy](#smaller-policy) applies.
+
+Note: you can always trigger a new transformation to be applied by using the `setContentSize` or `setContainerSize` APIs.
 
 #### Smaller policy
 
-You can also control how the content will be positioned when it is smaller than the container size through the `setSmallerPolicy(int)` method or by using the `smallerPolicy` attribute on the `ZoomLayout`.  By default, content will be centered when it is smaller than it's container.
+You can control how the content will be positioned when it is smaller than the container through 
+the `setSmallerPolicy(int)` method or by using the `smallerPolicy` XML attribute of `ZoomLayout` and `ZoomImageView`.
+By default, content is always centered when it is smaller than its container.
```
markusressel, line 178: The one thing I still find confusing when reading this documentation is **when** and **how often** this `smaller policy` applies. I thought (and the code also looks like it) that it will have an effect even after the initial base transformation is done (which is the first paragraph of the **Transformations** section) and also specifies the behavior when zooming out on a `contentView` that initially was bigger than the container.
I think this should be added to the documentation somewhere as the current text was not clearly explaining that and I again found myself digging through the code to understand how it is supposed to work.

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -638,42 +654,51 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
...
 
         val overScrollable = if (horizontal) mOverScrollHorizontal else mOverScrollVertical
         @ScaledPan val overScroll = (if (overScrollable && allowOverScroll) maxOverScroll else 0).toFloat()
-        return getPanCorrection(value, viewSize, contentSize, overScroll, basePanValue)
+        return getPanCorrection(value, containerSize, contentSize, overScroll, horizontal)
     }
 
     @ScaledPan
-    private fun getPanCorrection(@ScaledPan value: Float, viewSize: Float,
+    private fun getPanCorrection(@ScaledPan value: Float, containerSize: Float,
```
markusressel: while you're at it, could you add a documentation text to this method? It  is executed very frequently and takes a bunch of arguments that would benefit from additional documentation. Maybe the `checkPanBounds` method too.

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -638,42 +654,51 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
...
+                    min = correction
+                    max = correction
+                }
+                ZoomApi.SMALLER_POLICY_CENTER  -> {
+                    // Stay centered. Need a positive translation, that shows some background.
+                    min = extraSpace / 2f
+                    max = extraSpace / 2f
+                }
+                ZoomApi.SMALLER_POLICY_NONE -> {
+                    // Everything is fine, just don't exit the container boundaries.
```
markusressel, line 765: (No problem here, just a random tought while reading the comment :sweat_smile: )

markusressel (changes requested):
Otherwise the code looks fine.
I want to mention that I **have not tested this myself** but I am confident you did :+1:

README.md
natario1: I think I will open a new issue later to discuss this whole smaller policy thing before next release. We will improve docs after this is sorted out

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
natario1: I went on and removed the method :-)
Merged with another.

markusressel (approved):

PR #76, opened 2019-01-12
Feature/alignment

### LINKED ISSUE #75: Redefining SmallerPolicy
I would like to say what I have in mind about this new feature introduced in #71 that looks important to me and get feedback. This might also help in improving documentation.

## Transformations

As said elsewhere, _transformation_ is not the best name ever, but that's what we have.
The transformation defines the engine **resting position**. It is a keyframe that is reached at certain points, like at start up or when certain methods are called (`setContentSize` or `setContainerSize` with `applyTransformation = true`).

This was extremely useful anytime I had to use this library, plus it also defines the difference between `zoom` and `realZoom` (other names that I would change today). With transformations like `CENTER_CROP` or `CENTER_INSIDE`, we can assume that the content - **whatever its real size is** - will be fitting one of the container dimensions, and we call that `zoom == 1`. This is super useful for defining minZoom and maxZoom constraints. (`CENTER_CROP` and `minZoom = 1` make sure that you never show background, for example).

As it is defined now, transformation is a purely _zoom_ thing.

## Transformation Gravity

Some transformations / aspect ratios might leave with a situation where, after applying the transformation, part of the content is cropped. In these cases, the engine could align the content to different parts of the container, so the **transformation gravity** applies. This is a purely _pan_ thing.

**Error:** this gravity does not apply when the transformation leaves the content smaller than the container. Only the opposite. This feels wrong but also right, in that, the 'smaller policy' is a behavior policy that must apply always, not just at transformation time. This makes me think that this method should not exist at all.

## When content is smaller than container

This can happen in various situations:

- With `TRANSFORMATION_NONE` and a content that is actually smaller
- With `TRANSFORMATION_CENTER_INSIDE`, one of the axes will probably be smaller
- Unless minZoom is very high, pinching out can always make content smaller than the container

We've been ignoring this case until it was brought up in #71 . In this situation, I see that the engine might actually do two things:

- Let the content be moved around freely ( = `SMALLER_POLICY_NONE`)
- Align the content to some part of the container according to some gravity flag   

I can't think of any third option. If you can, please say. But if these are the only two options...

## Proposal: Alignment (Gravity)

Remove SmallerPolicies. Create `ALIGNMENT_` integer flags as always. Then create `alignment` XML attribute and `setAlignment` method.

Possible values should be 

- `ALIGNMENT_NONE` : corresponds to `SMALLER_POLICY_NONE`
- `ALIGNMENT_TOP`, `ALIGNMENT_BOTTOM`, ... : aligns to that specific side of the container

Docs should made clear that alignments only apply when content is smaller than the container, because forcing an alignment when content is bigger, makes the hidden content unreachable. This is actually obvious, not a design decision. When content is bigger you must be able to pan around.

#### About the name

I was also tempted to call this `setGravity`, I see arguments on both sides. Let me know what you think. For example (good and bad):

- `setGravity` makes confusion with `setTransformationGravity`
- `setGravity` is consistent with `setTransformationGravity`
- `setAlignment` being more generic might allow for more behaviors in the future
- The `gravity` XML attribute might be already used for some views?
- `setGravity` means we're using the Gravity constants. 
- At the same time this is confusing because we will also have GRAVITY_NONE and also we **do not** support all the gravity fields. A gravity `"top|bottom"` or `"fill_horizontal"` is ignored by the engine.
- With alignment we are free to do more fancy stuff. For example, `"none_horizontal|top"`, `"center_horizontal|none_vertical"`. We could have `none` as a flag and apply it to single axis.

I do prefer alignment.

#### About transformationGravity

After this change, docs should say this about transformations:

> When a transformation is applied, the content can be smaller or bigger than the container on some axis. When it is smaller, the `alignment` gravity applies as always. When it is bigger, the `alignment` gravity applies as well, **unless** a `transformationGravity` was defined, in which case it is respected instead.

So we should be inverting the dependency, make `transformationGravity` default to the `alignment` and not the opposite. Because `transformationGravity` is just a pan, one-time translation applied after the transformation and only when content is bigger. The `alignment` is a key behavior of the engine, always active.

What do you think about this?

@markusressel

natario1:
Fixes #75


README.md
```diff
@@ -142,46 +152,60 @@ There is no strict limit over what you can do with a `Matrix`,
...
 |`centerInside`|The content is scaled down or up so that it fits completely inside the view bounds.|
 |`centerCrop`|The content is scaled down or up so that its smaller side fits exactly inside the view bounds. The larger side will be cropped.|
 |`none`|No transformation is applied.|
 
-The engine applies the given transformation, and any minZoom and maxZoom constraints.
-
-If, after this process, the content is bigger than the container, the engine will also apply a
-translation according to the given transformation gravity.
+After transformation is applied, the transformation gravity will reposition the content with
+the specified value. Supported values are most of the Android `Gravity` flags, plus `TRANSFORMATION_GRAVITY_AUTO`.
```
markusressel: I think it is better to add an example so people know exactly what constants you are talking about, like:

> Supported values are most of the Android `android.view.Gravity` flags like `Gravity.TOP`, plus `TRANSFORMATION_GRAVITY_AUTO`.

README.md
```diff
@@ -142,46 +152,60 @@ There is no strict limit over what you can do with a `Matrix`,
...
+|`auto` (default)|The transformation gravity is taken from the engine [alignment](#alignment), defaults to `center` on both axes.|
 
-If, after this process, the content is smaller than the container, note that the current
-[Smaller Policy](#smaller-policy) applies.
+**Note: after transformation and gravity are applied, the engine will apply - as always - all the active constraints,
+including minZoom, maxZoom, alignment. This means that the final position might be slightly (or completely) different.**
 
-Note: you can always trigger a new transformation to be applied by using the `setContentSize` or `setContainerSize` APIs.
+For example, when `maxZoom == 1`, is forced to not be any larger than the container. This means that
+a `centerCrop` transformation will not work (will act just like a `centerInside`).
```
markusressel: Well it will work, but it will have the same effect as `centerInside` so it might not be the intended behavior.

README.md
```diff
@@ -142,46 +152,60 @@ There is no strict limit over what you can do with a `Matrix`,
...
+|`top`, ...|The content is panned so that its *top* side matches teh container *top* side. Same for other values.|
+|`auto` (default)|The transformation gravity is taken from the engine [alignment](#alignment), defaults to `center` on both axes.|
 
-If, after this process, the content is smaller than the container, note that the current
-[Smaller Policy](#smaller-policy) applies.
+**Note: after transformation and gravity are applied, the engine will apply - as always - all the active constraints,
+including minZoom, maxZoom, alignment. This means that the final position might be slightly (or completely) different.**
 
-Note: you can always trigger a new transformation to be applied by using the `setContentSize` or `setContainerSize` APIs.
+For example, when `maxZoom == 1`, is forced to not be any larger than the container. This means that
```
markusressel: This sentence doesn't seem to make sense. Did you mean:

> This can happen for example, when you set `maxZoom == 1`, so the content is forced to not be any larger than the container.

README.md
```diff
@@ -142,46 +152,60 @@ There is no strict limit over what you can do with a `Matrix`,
...
+or the `alignment` XML flag of `ZoomLayout` and `ZoomImageView`. 
+The default value is `Alignment.CENTER` which will center the content on both directions.
 
-|Policy|Description|
-|------|-----------|
-|`center`|The content will be centered within the container.|
-|`fromTransformation`|The content will respect the gravity parameter of the transformation (which defaults to `Gravity.CENTER` as well).|
-|`none`|The content is free to be moved around inside the container bounds.|
+**Note: alignment does not make sense when content is larger than the container, because forcing an 
+alignment (e.g. left) would mean making part of the content unreachable (e.g. the right part).**
```
markusressel, line 194: [...] so it is only applied when the content is smaller than the container (e.g. `zoom < 1`)

README.md
```diff
@@ -142,46 +152,60 @@ There is no strict limit over what you can do with a `Matrix`,
...
+
+|Alignment|Description|
+|---------|-----------|
+|`top`, `bottom`, `left`, `right`|Force align the content to the same side of the container.|
+|`center_horizontal`, `center_vertical`|Force the content to be centered inside the container on that axis.|
+|`none_horizontal`, `none_vertical`|No alignment set: content is free to be moved on that axis.|
```
markusressel, line 200: Are you going with those constants nonetheless?
Ah so they will be used internally but the dev user doesn't have to specify them. :+1:

README.md
```diff
@@ -142,46 +152,60 @@ There is no strict limit over what you can do with a `Matrix`,
...
+
+You can use the or operation to mix the vertical and horizontal flags.
```
markusressel: What? ^^
Ahh you mean the `or` operator! Those backticks are crucial here xD

library/src/main/java/com/otaliastudios/zoom/Alignment.kt
```diff
@@ -0,0 +1,50 @@
+package com.otaliastudios.zoom
+
+object Alignment {
```
markusressel, line 9: missing documentation on class

library/src/main/java/com/otaliastudios/zoom/ZoomApi.kt
```diff
@@ -114,18 +114,30 @@ interface ZoomApi {
...
+        com.otaliastudios.zoom.Alignment.BOTTOM,
+        com.otaliastudios.zoom.Alignment.CENTER_VERTICAL,
+        com.otaliastudios.zoom.Alignment.NONE_VERTICAL,
+        com.otaliastudios.zoom.Alignment.TOP,
+        com.otaliastudios.zoom.Alignment.LEFT,
+        com.otaliastudios.zoom.Alignment.CENTER_HORIZONTAL,
+        com.otaliastudios.zoom.Alignment.NONE_HORIZONTAL,
+        com.otaliastudios.zoom.Alignment.RIGHT,
+        com.otaliastudios.zoom.Alignment.CENTER,
+        com.otaliastudios.zoom.Alignment.NONE,
```
markusressel, line 137: I think the `Alignment` class could be imported, or is there another `Alignment` import already? Would look much cleaner imo.

markusressel: Oh I see it's because the annotation class is named Alignment too right?

library/src/main/java/com/otaliastudios/zoom/ZoomApi.kt
```diff
@@ -189,6 +201,16 @@ interface ZoomApi {
...
     fun setAllowFlingInOverscroll(allow: Boolean)
 
+    /**
+     * Sets the base transformation to be applied to the content.
+     * See [setTransformation].
+     *
+     * @param transformation the transformation type
+     */
+    fun setTransformation(@Transformation transformation: Int) {
+        setTransformation(transformation, TRANSFORMATION_GRAVITY_AUTO)
```
markusressel, line 211: I think this can be replaced with a default value on `fun setTransformation(@Transformation transformation: Int, gravity: Int = TRANSFORMATION_GRAVITY_AUTO)` and a `@JvmOverloads` annotation.

library/src/main/java/com/otaliastudios/zoom/ZoomApi.kt
```diff
@@ -200,13 +222,16 @@ interface ZoomApi {
     fun setTransformation(@Transformation transformation: Int, gravity: Int)
 
     /**
-     * Sets the policy to use when content is smaller than the container.
-     * Defaults to [SMALLER_POLICY_CENTER]
-     * which means that the content will be centered.
+     * Sets the content alignment. Can be any of the constants defined in [com.otaliastudios.zoom.Alignment].
```
markusressel, line 225: Is it really necessary to mention the `com.otaliastudios.zoom.Alignment` class? The `@Annotation` annotation on the parameter should already specify this and if I'm not mistaking the IDE will correct you (the dev using this library) too.

library/src/main/java/com/otaliastudios/zoom/ZoomApi.kt
```diff
@@ -200,13 +222,16 @@ interface ZoomApi {
...
+     * The content will be aligned and forced to the specified side of the container.
+     * Defaults to [com.otaliastudios.zoom.Alignment.CENTER].
      *
-     * @param policy the policy
+     * Of course, this is disabled when the content is larger than the container,
```
markusressel: > Of course, [...]

I don't like this phrasing... I would go with:
>  Keep in mind that this is disabled when the content is larger than the container, because a forced alignment at all times would result in part of the content being unreachable.
But do as you like.

library/src/main/java/com/otaliastudios/zoom/ZoomApi.kt
```diff
@@ -354,23 +379,35 @@ interface ZoomApi {
         const val TRANSFORMATION_NONE = 2
 
         /**
-         * Constant for [ZoomApi.setSmallerPolicy].
-         * When smaller than its container, the content will be centered.
+         * Constant for [ZoomApi.setTransformation] gravity.
+         * This currently means that the gravity will be inferred from the alignment or
```
markusressel: currently?
This implies that this will change in the short term future. Will it?

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -428,8 +428,18 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
         mTransformationGravity = gravity
     }
 
-    override fun setSmallerPolicy(@SmallerPolicy policy: Int) {
-        mSmallerPolicy = policy
+    /**
+     * Sets the content alignment. Can be any of the constants defined in [Alignment].
+     * The content will be aligned and forced to the specified side of the container.
+     * Defaults to [Alignment.CENTER].
```
markusressel: Well `ZoomApi.ALIGNMENT_DEFAULT` which is set to `Alignment.CENTER`... don't really know if that's important but if it's ever going to change it will likely be forgotten to be changed in this documentation part.

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -428,8 +428,18 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
...
+     *
+     * Of course, this is disabled when the content is larger than the container,
+     * because a forced alignment would mean making part of the content unreachable.
```
markusressel: Same thing on the phrasing here.

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -428,8 +428,18 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
...
+     *
+     * @param alignment the new alignment
+     */
+    override fun setAlignment(alignment: Int) {
```
markusressel: Is the `@Alignment` annotation missing here?

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -613,26 +623,47 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
...
+            }
+            val vertical = when (verticalAlignment) {
+                Alignment.TOP -> Gravity.TOP
+                Alignment.BOTTOM -> Gravity.BOTTOM
+                Alignment.CENTER_VERTICAL, Alignment.NONE_VERTICAL -> Gravity.CENTER_VERTICAL
+                else -> Gravity.CENTER_VERTICAL
+            }
+            return horizontal or vertical
+        }
+    }
```
markusressel, line 654:
```
return when (input) {
    ZoomApi.TRANSFORMATION_GRAVITY_AUTO -> {
        val horizontalAlignment = Alignment.getHorizontal(mAlignment)
        val verticalAlignment = Alignment.getVertical(mAlignment)
        val horizontal = when (horizontalAlignment) {
            Alignment.LEFT -> Gravity.LEFT
            Alignment.RIGHT -> Gravity.RIGHT
            Alignment.CENTER_HORIZONTAL, Alignment.NONE_HORIZONTAL -> Gravity.CENTER_HORIZONTAL
            else -> Gravity.CENTER_HORIZONTAL
        }
        val vertical = when (verticalAlignment) {
            Alignment.TOP -> Gravity.TOP
            Alignment.BOTTOM -> Gravity.BOTTOM
            Alignment.CENTER_VERTICAL, Alignment.NONE_VERTICAL -> Gravity.CENTER_VERTICAL
            else -> Gravity.CENTER_VERTICAL
        }
        horizontal or vertical
    }
    else -> input
}
```

I guess it's personal preference, but I just love `when` :smile:

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -735,38 +766,38 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
...
-                else -> throw IllegalStateException("Unsupported policy: $mSmallerPolicy")
+            val correction = when (alignment) {
+                Alignment.TOP -> applyGravity(Gravity.TOP, extraSpace, horizontal)
+                Alignment.BOTTOM -> applyGravity(Gravity.BOTTOM, extraSpace, horizontal)
+                Alignment.LEFT -> applyGravity(Gravity.LEFT, extraSpace, horizontal)
+                Alignment.RIGHT -> applyGravity(Gravity.RIGHT, extraSpace, horizontal)
+                Alignment.CENTER_VERTICAL -> applyGravity(Gravity.CENTER_VERTICAL, extraSpace, horizontal)
+                Alignment.CENTER_HORIZONTAL -> applyGravity(Gravity.CENTER_HORIZONTAL, extraSpace, horizontal)
+                else -> null
+            }
```
markusressel: This is a lot of code duplication, maybe extract the `applyGravity()` part like this?:

```
val gravity = when (alignment) {
    Alignment.TOP -> Gravity.TOP
    Alignment.BOTTOM -> Gravity.BOTTOM
    Alignment.LEFT -> Gravity.LEFT
    Alignment.RIGHT -> Gravity.RIGHT
    Alignment.CENTER_VERTICAL -> Gravity.CENTER_VERTICAL
    Alignment.CENTER_HORIZONTAL -> Gravity.CENTER_HORIZONTAL
    else -> null
}

val correction = if (gravity != null) applyGravity(gravity, extraSpace, horizontal) else null
if (correction != null) {
[...]
```

Still looks not that great though.

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -1502,32 +1533,38 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
...
+        val alignment = if (horizontal) Alignment.getHorizontal(mAlignment) else Alignment.getVertical(mAlignment)
+        if (contentDim > containerDim) {
+            // Content is bigger. We can move between 0 and extraSpace, but since our pans
+            // are negative, we must invert the sign.
+            val extraSpace = contentDim - containerDim
+            output.minValue = -extraSpace
             output.maxValue = 0
+        } else if (alignment == Alignment.NO_VALUE
+                || alignment == Alignment.NONE_VERTICAL
+                || alignment == Alignment.NONE_HORIZONTAL) {
```
markusressel: This should also work:
`else if (alignment in setOf(Alignment.NO_VALUE, Alignment.NONE_VERTICAL, Alignment.NONE_HORIZONTAL)) {`
I don't know about performance though.

markusressel (changes requested):
Great work!
Again just fyi: I did **not** test this in any way.

README.md
natario1: Yes, we need them internally so there's no point in hiding them. However user doesn't have to specify explicitly

README.md
natario1: I will address all these in a while. But wanted to say that `zoom < 1` does not necessarily mean that content is smaller than container. It does only with CENTER_CROP or CENTER_INSIDE transformations and even then, it depends on the axis.

Just saying because I have seen this check (zoom < 1F) in the pivot function, and I don't know if it's correct or not. I'm tempted to say that it's not, and I also wonder if that should be changed according to the alignment? Not sure.

library/src/main/java/com/otaliastudios/zoom/ZoomApi.kt
natario1: Yes! But this won't be a problem for the user as no-one will likely import the `Alignment` annotation.

library/src/main/java/com/otaliastudios/zoom/ZoomApi.kt
natario1: It's not possible on interfaces unfortunately

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
natario1: I think it looks clean now!

README.md
markusressel: Hmm yes that makes sense. I guess I was lucky in my testing.

Yes I think so too. As i mentioned in the comment there I initially wanted to base it on the transformationGravity but as the alignment will be the new thing after this PR it should be based on it instead.

library/src/main/java/com/otaliastudios/zoom/ZoomApi.kt
markusressel: Ah didn't think about that. Sad :disappointed:

README.md
natario1: Can you explain what that function should return? It's not 100% clear to me, I didn't investigate, but I can try to update

markusressel: Well the function decides what pivot point should be used for the zoom animation based on what corner of the screen is currently overscrolled (or will be, with the target zoom applied). I did this because applying the pan fix and the zoom fix at the same time in an animation when overscrolled while zoomed in a bit (specifically overpinching and overscrolling at the same time in a single pinch gesture) didn't result in the expected target position and/or a funky animation. Implementing this was a bit of trial and error tbh and I think this method is only applicable when the content is zoomed in meaning that **at most** two sides of the content are overscrolled/have visible gray areas. Because of that I added the `zoom < 1` condition which just takes the center of the screen as the pivot point when the view has more than two gray areas around it (which doesn't seem to be the correct way of figuring that out). My current guess would be that this should be based on the alignment instead of just using the center when there are more than two gray areas but this whole method (and it's purpose) may require a complete rethinking.

natario1: I understand. Looks like it might take some time, I think I will merge this now and we'll fix later.

PR #85, opened 2019-02-06
Cancel active fling animation when appropriate

### LINKED ISSUE #84: Fling animation overrides `moveTo` calls
When a fling animation is in progress all calls to `moveTo` (and similar methods without animation I guess) are executed but useless since the position will instantly be changed again by the active fling animation.

I don't know what the best solution to this is. I think there are two options:
* automatically stop any fling when one of the methods for changing pan is called
* provide a method (f.ex. `cancelFling`) to let the dev decide if a fling should be cancelled

or maybe even both?

@natario1 thoughts on that?

markusressel:
added cancelFling method to ZoomApi
cancel fling in moveTo, panBy and zoomTo if necessary

fixes #84

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,7 +268,7 @@ In any case the current scale is not considered, so your system won't change if 
 |`setAllowFlingInOverscroll(boolean)`|If true, fling gestures will be allowed even when detected while overscrolled. This might cause artifacts so it is disabled by default.|`false`|
 |`panTo(float, float, boolean)`|Pans to the given values, animating if needed.|`-`|
 |`panBy(float, float, boolean)`|Applies the given deltas to the current pan, animating if needed.|`-`|
-|`cancelAnimations()`|Cancels all currently active code driven animations (including fling animations).|`-`|
+|`cancelAnimations()`|Cancels all currently active animations triggered by either API calls with `animate = true` or touch input flings.|`-`|
 
 The `moveTo(float, float, float, boolean)` API will let you animate both [zoom](#zoom) and pan at the same time.
 
--- a/README.md
+++ b/README.md
@@ -243,6 +243,7 @@ will make more sense than the other - e. g., in a PDF viewer, you might want to 
 |`zoomOut()`|Applies a small, animated zoom-out.|`-`|
 |`setZoomEnabled(boolean)`|If true, the content will be allowed to zoom in and out by user input.|`true`|
 
+
 The `moveTo(float, float, float, boolean)` API will let you animate both zoom and [pan](#pan) at the same time.
 
 ### Pan
@@ -267,6 +268,7 @@ In any case the current scale is not considered, so your system won't change if 
 |`setAllowFlingInOverscroll(boolean)`|If true, fling gestures will be allowed even when detected while overscrolled. This might cause artifacts so it is disabled by default.|`false`|
 |`panTo(float, float, boolean)`|Pans to the given values, animating if needed.|`-`|
 |`panBy(float, float, boolean)`|Applies the given deltas to the current pan, animating if needed.|`-`|
+|`cancelAnimations()`|Cancels all currently active code driven animations (including fling animations).|`-`|
 
 The `moveTo(float, float, float, boolean)` API will let you animate both [zoom](#zoom) and pan at the same time.
 
--- a/library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
+++ b/library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
@@ -85,7 +85,6 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
     private var mInitialized = false
     private var mContentScaledRect = RectF()
     private var mContentRect = RectF()
-    private var mClearAnimation = false
     private var mAnimationDuration = DEFAULT_ANIMATION_DURATION
 
     // Gestures
@@ -677,27 +676,31 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
 
     // Returns true if we should go to that mode.
     @SuppressLint("SwitchIntDef")
-    private fun setState(@State state: Int): Boolean {
-        LOG.v("trySetState:", state.toStateName())
+    private fun setState(@State newState: Int): Boolean {
+        LOG.v("trySetState:", newState.toStateName())
         if (!mInitialized) return false
-        if (state == mState) return true
-        val oldMode = mState
+        // we need to do some cleanup in case of ANIMATING so we can't return just yet
+        if (newState == mState && newState != ANIMATING) return true
+        val oldState = mState
 
-        when (state) {
-            SCROLLING -> if (oldMode == PINCHING || oldMode == ANIMATING) return false
-            FLINGING -> if (oldMode == ANIMATING) return false
-            PINCHING -> if (oldMode == ANIMATING) return false
+        when (newState) {
+            SCROLLING -> if (oldState == PINCHING || oldState == ANIMATING) return false
+            FLINGING -> if (oldState == ANIMATING) return false
+            PINCHING -> if (oldState == ANIMATING) return false
             NONE -> dispatchOnIdle()
         }
 
         // Now that it succeeded, do some cleanup.
-        when (oldMode) {
+        when (oldState) {
+            ANIMATING -> {
+                mActiveAnimators.forEach { it.cancel() }
+                mActiveAnimators.clear()
+            }
             FLINGING -> mFlingScroller.forceFinished(true)
-            ANIMATING -> mClearAnimation = true
         }
 
-        LOG.i("setState:", state.toStateName())
-        mState = state
+        LOG.i("setState:", newState.toStateName())
+        mState = newState
         return true
     }
 
@@ -1202,17 +1205,13 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
      * Applies a small, animated zoom-in.
      * Shorthand for [zoomBy] with factor 1.3.
      */
-    override fun zoomIn() {
-        zoomBy(1.3f, animate = true)
-    }
+    override fun zoomIn() = zoomBy(1.3f, animate = true)
 
     /**
      * Applies a small, animated zoom-out.
      * Shorthand for [zoomBy] with factor 0.7.
      */
-    override fun zoomOut() {
-        zoomBy(0.7f, animate = true)
-    }
+    override fun zoomOut() = zoomBy(0.7f, animate = true)
 
     /**
      * Animates the actual matrix zoom to the given value.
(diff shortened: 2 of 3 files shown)


natario1:
Cool! Can you call this cancel() and cancel both STATE_FLINGING and STATE_ANIMATING ? Would this fit your requirements? Docs would say that this cancels any animated motion like flings or developer started animations, while of course pinch/scrolls can't be canceled because the finger is still there.

Not against a more specific API but I think that all these names (scroll/pan/pinch/zoom/fling) in API and XML attributes are already confusing enough. A simple cancel() or cancelAnimations() would be better.

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -822,6 +822,9 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
         LOG.v("processTouchEvent:", "start.")
         if (mState == ANIMATING) return TOUCH_STEAL
 
+        // if we are flinging, just cancel it and process touch as normal
+        cancelFling()
```
natario1: I don't think this is needed as as soon as we find a new gesture, that will cancel the fling through `setState()`

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -1128,6 +1131,7 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
      */
     override fun moveTo(@Zoom zoom: Float, @AbsolutePan x: Float, @AbsolutePan y: Float, animate: Boolean) {
         if (!mInitialized) return
+        cancelFling()
```
natario1: Not needed in the `animate` case, as `setState()` there will cancel what's needed.

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -1163,6 +1167,7 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
      */
     override fun panBy(@AbsolutePan dx: Float, @AbsolutePan dy: Float, animate: Boolean) {
         if (!mInitialized) return
+        cancelFling()
```
natario1: Same as above

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -1179,6 +1184,7 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
      */
     override fun zoomTo(@Zoom zoom: Float, animate: Boolean) {
         if (!mInitialized) return
+        cancelFling()
```
natario1: Same as above. Correct me if I'm wrong

natario1:
Looking at the code now, I would also appreciate if you could simplify and remove the `mClearAnimation` logic and instead rely to `setState()`.

I mean that instead of checking `mClearAnimation == true`, the animation APIs could check that `state` is still `STATE_ANIMATING`. If state changed, the animation must stop. It makes more sense, as we don't want a clear flag for every kind of motion. `setState` seems the right place to coordinate.

markusressel:
It's funny because at first I also created a cancel flag for the fling :smile: But then I realized the state is also reflecting this and was used already to stop the fling in some situations.

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
markusressel: Adding this will also cancel the fling when the user just touches the screen though. I don't think that's currently the case. AFAIK currently you can't stop a fling by any means. you can only override it with a new fling (with a new gesture as you said) or by using a method with animation (which will change the state and stop the fling through that). It should be possible to fling and stop at any given point though. The only problem with this is that a short tap on the screen will not only stop the fling but also register as a "click" which is not desirable. I will check on that.

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
markusressel: Correct, I only did it to make sure it is always stopped... but I guess it doesn't add much to double cancel it.

markusressel:
When I think about it, canceling ongoing animations from the update method isn't a good idea anyway because it might be possible to switch to a new animation while an old animation is still around. In that case the update method will certainly almost never reach it's `if` line to stop and both animations will run at the same time.

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
markusressel: I think I got a little excited about the idea of supporting this behaviour with just the addition of a single line. I spend the last hours trying to figure this out...

Imo there should be only a single code section where the "fix animation" is started, and that is when the user releases touch completely. I tried to implement this but had lots of strange behavior with the animation never triggering or `NONE` states right after the animation started and even if that worked the behavior changed when an animation was already running when touching the screen. I also had a hard time understanding how and why return values of the pinch and flingscroll listener are used in the `processTouchEvent` method and how they affect the overscroll/pan fix animation.

My final resolution is that this is a topic in itself and I will not attempt to implement it in this PR. I will rework this PR and force push anything unrelated - but I have to give my brain a little rest know :smile: I would really like this to happen though as it is very unintuitive that flings and animations can not be cancelled by user touch.
I will inform you if I am ready for a new review.

natario1: Why is that? Note that there is also a `setState(NONE)` in `onDown()`, which was put there to do this exact job. And might be causing some of the weirdness you have been experiencing.

Without trying anything I would say that that specific `setState(NONE)` is enough to cancel animations on user touch:

- Developer animations: if you replace `mCancelAnimations == true` with `mState == ANIMATING`, this check will fail at the next cycle, so that's fine. It will stop.

- Fling animations: I don't know how it works at the moment, but the flinger could also be doing the same - check that state is still FLINGING before applying the new values at each step.

These are just a few lines. Then on top of this we could also have the external API `cancelAnimations()` which implementation is basically set state to NONE if it was FLINGING or ANIMATING.

The `processTouchEvent` part is delicate indeed because it was just a lot of trial and error to get it right. I wouldn't touch that too much, I'm not surprised that you've seen weird stuff :) see if the setState during onDown is enough

natario1: If after doing these simple things (stop animations and flings by checking `state` in their own cycles), animations are still not stopped on touch... then I would like to understand why, it's very strange.

natario1:
> When I think about it, canceling ongoing animations from the update method isn't a good idea anyway because it might be possible to switch to a new animation while an old animation is still around. In that case the update method will certainly almost never reach it's `if` line to stop and both animations will run at the same time.

Sorry, I understand this comment now!

Need an extra check then... Maybe we could keep an `animationCount` integer, such that each animation that is started does `val thisCount = animationCount++`. Then at its own cycle we can check that `state == STATE_ANIMATING && thisCount == animationCount`. What do you think?

Flings should not have this issue as they are started by input events and you can't have two at the same time

markusressel:
> Need an extra check then... Maybe we could keep an `animationCount` integer, such that each animation that is started does `val thisCount = animationCount++`. Then at its own cycle we can check that `state == STATE_ANIMATING && thisCount == animationCount`. What do you think?
> 
> Flings should not have this issue as they are started by input events and you can't have two at the same time

I would just keep a reference of all active animators and cancel them if necessary.
Take a look at the current approach and let me know if you see problems with it.

**EDIT:**
Adding on the "you can't have two at the same time" thing, thats what I expected to be the case with animations too although one could argue that using pan and zoom animations at the same time can be a use case - but then again the dev would have to specify that his other animation should not get killed because it would in the state proposed by this PR :confused:

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
markusressel: > Why is that? Note that there is also a `setState(NONE)` in `onDown()`, which was put there to do this exact job. And might be causing some of the weirdness you have been experiencing.

There is no `setState()` in the `onDown()` method of the `FlingScrollListener` if you mean that. And even if it was it wouldn't work because when an animation is running **every** touch input event is just discarded (first line in `processTouchEvent`).

> Without trying anything I would say that that specific `setState(NONE)` is enough to cancel animations on user touch:
> 
> * Developer animations: if you replace `mCancelAnimations == true` with `mState == ANIMATING`, this check will fail at the next cycle, so that's fine. It will stop.
> * Fling animations: I don't know how it works at the moment, but the flinger could also be doing the same - check that state is still FLINGING before applying the new values at each step.

I kind of did this, but in a different way (see other comment)
The `flingScroller` is already cancelled in `setState()`when the state changes and also works when a new fling is applied while an old fling is still animating (at least in my testing).

> These are just a few lines. Then on top of this we could also have the external API `cancelAnimations()` which implementation is basically set state to NONE if it was FLINGING or ANIMATING.

Yes I did it in exactly this way.

> The `processTouchEvent` part is delicate indeed because it was just a lot of trial and error to get it right. I wouldn't touch that too much, I'm not surprised that you've seen weird stuff :) see if the setState during onDown is enough

Well I'm glad I'm not the only one frustrated with this method then :smile:
Sadly it won't be that easy though as mentioned above.

markusressel:
I created a new issue (#86) for the touch input stopping animations thing.
Let's continue the discussion over there.

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -1336,23 +1351,22 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
...
             @Zoom val startZoom = this.zoom
             @Zoom val endZoom = checkZoomBounds(zoom, allowOverScroll)
             val startPan = pan
             val targetPan = AbsolutePoint(x, y)
             LOG.i("animateZoomAndAbsolutePan:", "starting.", "startX:", startPan.x, "endX:", x, "startY:", startPan.y, "endY:", y)
             LOG.i("animateZoomAndAbsolutePan:", "starting.", "startZoom:", startZoom, "endZoom:", endZoom)
 
-            @SuppressLint("ObjectAnimatorBinding")
-            val animator = ObjectAnimator.ofPropertyValuesHolder(mContainer,
+            ObjectAnimator.ofPropertyValuesHolder(mContainer,
```
natario1: Can you remove mContainer from this constructor?

As far as I remember, It might be useless and harmful. I think ObjectAnimator will try to call `mContainer.setPan()` and `mContainer.setZoom()` with reflection (failing to do so).

README.md
```diff
@@ -267,6 +268,7 @@ In any case the current scale is not considered, so your system won't change if
 |`setAllowFlingInOverscroll(boolean)`|If true, fling gestures will be allowed even when detected while overscrolled. This might cause artifacts so it is disabled by default.|`false`|
 |`panTo(float, float, boolean)`|Pans to the given values, animating if needed.|`-`|
 |`panBy(float, float, boolean)`|Applies the given deltas to the current pan, animating if needed.|`-`|
+|`cancelAnimations()`|Cancels all currently active code driven animations (including fling animations).|`-`|
```
natario1: I would say a finger fling is not code driven?

natario1:
> Adding on the "you can't have two at the same time" thing, thats what I expected to be the case with animations too although one could argue that using pan and zoom animations at the same time can be a use case - but then again the dev would have to specify that his other animation should not get killed because it would in the state proposed by this PR 😕

I don't have a strong position about this, I'm OK with merging your autocancel, but yes, in theory we could avoid. If you run two animations, the second should win in the end - not sure about the eye result, but I think it should run after the first at each cycle, invalidating the first. You just get a mess of listener updates. For this the dev could call `cancelAnimations()`.

Buut again, not sure :-) 

Should we add an API like allowConcurrentAnimations ?

Or, in the long run, have two booleans `stopConcurrentAnimations` and `stopAnimationsOnTap`.

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
markusressel: Ohh so I can just remove the parameter completely! Yes that makes sense.
I used this kind of "multiple value animations" animator the first time there. Didn't think it though back then.

markusressel:
> You just get a mess of listener updates.

You're right, this would be impossible to handle properly if the dev is allowed to start multiple animations at the same time... I don't think this is a good thing to allow then. The only drawback with limiting to a single animation concurrently is that f.ex. a `moveTo` call with another `moveTo` call in the middle will cause an abrupt stop of the animation before the new animation gets going. But first I think this is a minimal issue and second it should also be handled by the library (if ever).

> Should we add an API like allowConcurrentAnimations ?

I don't think so (based on the text above).

README.md
markusressel: Well the user isn't touching the screen anymore so... I would argue it is :smile: but I see your point.
Do you have a suggestion? Would you just write:
```
Cancels all currently active animations.
```
?

natario1: I would be more specific and say like, animations triggered by API calls with `animate = true` and fling animations triggered by touch input.

markusressel: @natario1 Ok with the current vesion?

natario1: Looks good to me! @markusressel feel free to create a new release if you think so

natario1 (approved):

PR #90, opened 2019-02-26
Check vertical alignment in toVerticalGravity (instead of horizontal).

asclepix:
I've noticed differnt behaviour horizontally and vertically of the Alignment property and I think this is the cause.


natario1:
Makes sense! Thank you very much 👍

natario1 (approved):

PR #92, opened 2019-02-28
ZoomLayout implements OnGlobalLayoutListener

### LINKED ISSUE #58: Unregister ZoomLayout's OnGlobalLayoutListener

dmazzoni:
Fixes #58, improving on #65.

### DIFF
--- a/library/src/main/java/com/otaliastudios/zoom/ZoomLayout.kt
+++ b/library/src/main/java/com/otaliastudios/zoom/ZoomLayout.kt
@@ -4,10 +4,7 @@ import android.content.Context
 import android.graphics.Canvas
 import android.graphics.Matrix
 import android.util.AttributeSet
-import android.view.Gravity
-import android.view.MotionEvent
-import android.view.View
-import android.view.ViewGroup
+import android.view.*
 import android.widget.FrameLayout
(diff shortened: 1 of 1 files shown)


natario1 (approved):

natario1:
Thank you for spotting this @dmazzoni !

PR #94, opened 2019-03-02
ZoomSurfaceView

### LINKED ISSUE #88: Create ZoomSurfaceView
I have tried in many ways to use SurfaceView with ZoomLayout, I would like to know if it is possible to use it, thanks.

natario1:
This brings in a third piece called `ZoomSurfaceView` which is a zoom-powered `GLSurfaceView`.

Eventually I plan to move the `.opengl.` stuff to a separate library since I use these quite often. This lib will depend on that one.

Fixes #88


natario1:
@markusressel any chance you could run the demo app and take a look at the scale issue. I can't get why it doesn't work as expected, seems like the pivot point is incorrect.

I think there's some logic error in `ZoomSurfaceView.onDrawFrame` (which by the way is the only relevant class, the rest in the `opengl` package are just helpers and I know they work well).

markusressel:
> @markusressel any chance you could run the demo app and take a look at the scale issue. I can't get why it doesn't work as expected, seems like the pivot point is incorrect.
> 
> I think there's some logic error in `ZoomSurfaceView.onDrawFrame` (which by the way is the only relevant class, the rest in the `opengl` package are just helpers and I know they work well).

Sure I can take a look, I am currently quite busy though so I don't think I will get to this before the next weekend.

natario1:
Cool, thanks! I will give some more thought tomorrow also and see if I can figure it out

natario1:
It took some hours but I got it working in the end.

markusressel (approved):
Looks great so far!
I think extracting all those GL helper classes into a separate library would be nice though before merging this to keep git history clean if you want to do this anyway.

natario1:
It will have the extra benefit that we can require people to explicitly include that library, so this lib size is small if you don't want to use ZoomSurfaceView.

However it is really time consuming so I'm not sure when I will be able to do it.

markusressel (approved):

natario1:
Will merge this one!

PR #97, opened 2019-03-20
Split into components

natario1:
Splits the `ZoomEngine` into smaller components so it's easier to maintain and testable.

This also exposes `scaledPan`, `scaledPanX` and `scaledPanY` as they can be useful sometimes.


library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -223,44 +175,28 @@ internal constructor(context: Context) : ViewTreeObserver.OnGlobalLayoutListener
...
 
-    /**
-     * The current horizontal scaled pan, which is the pan position of the content
-     * according to the current zoom value (so it's scaled).
-     */
+    // TODO needed? access through MatrixManager?
     @ScaledPan
-    private val scaledPanX: Float
-        get() = mContentScaledRect.left
+    internal val contentScaledWidth get() = matrixManager.contentScaledWidth
```
markusressel: I don't know what your goal is with the matrix manager, but delegate methods like this could also be added implicitly  by using an interface (just like it is currently the case with `ZoomApi`).

If you want devs to use matrix manager instead, I would just remove those delegates.
(I have not thought about either solution in depth)

library/src/main/java/com/otaliastudios/zoom/internal/Controller.kt
```diff
@@ -0,0 +1,148 @@
...
+import android.view.MotionEvent
+import androidx.annotation.IntDef
+import com.otaliastudios.zoom.ZoomLogger
+
+/**
+ * Deals with touch input, holds the internal [state] integer,
+ * and applies special logic to touch inputs and state changes to
+ * prefer one gesture over the other, for example.
+ */
+internal class Controller(private val callback: Callback) {
```
markusressel: Imho the name `Controller` is way to generic. This controller could do literally anything.
It looks like it could be a `TouchInputController`?
Or maybe just an `InputController`?
Or simply a `ZoomEngineController`?

library/src/main/java/com/otaliastudios/zoom/internal/Dispatcher.kt
```diff
@@ -0,0 +1,42 @@
+package com.otaliastudios.zoom.internal
+
+import com.otaliastudios.zoom.ZoomApi
+import com.otaliastudios.zoom.ZoomEngine
+import com.otaliastudios.zoom.ZoomEngine.Listener
+
+internal class Dispatcher(private val engine: ZoomEngine) {
```
markusressel: I feel the same about this class name. What is it dispatching to and from whom and why?
There should be at least class documentation but I would also add docs to *any* method no matter the visibility.

library/src/main/java/com/otaliastudios/zoom/internal/MatrixManager.kt
```diff
@@ -0,0 +1,445 @@
...
+ */
+internal class MatrixManager(
+        private val zoomManager: ZoomManager,
+        private val panManager: PanManager,
+        private val controller: Controller,
+        private val callback: Callback,
+        private val engine: ZoomEngine
+) {
+
+    internal interface Callback {
```
markusressel: I personally don't really like these inner classes (or interfaces) and would rather use a separate file for this interface and give it a proper name, but I guess it's personal preference so I would leave it up to you.

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
natario1: It's just a matter of breaking the ZoomEngine monolith into testable components. In the future will create base interfaces for this components, so we can write tests.

I also wanted to bring this in before working on a new feature (rotation). I will refine this further and add docs 👍 

I wonder if, instead of keeping a `zoom` float, we could compute the current zoom by comparing `contentRect` and `contentScaledRect` (which we already use to compute pan), it would be much cleaner

natario1:
I'm satisfied now, what do you think? I have split things further and isolated features.

For example, now only ZoomEngine and ZoomManager know about the transformation. All the other components and APIs will use `@RealZoom` (which I plan to rename `@Zoom` one day and should be the default).

I have also done what I said in previous comment - we don't store a `zoom` float anymore, all we need are the two rects.

markusressel:
> I'm satisfied now, what do you think? I have split things further and isolated features.
> 
> For example, now only ZoomEngine and ZoomManager know about the transformation. All the other components and APIs will use `@RealZoom` (which I plan to rename `@Zoom` one day and should be the default).
> 
> I have also done what I said in previous comment - we don't store a `zoom` float anymore, all we need are the two rects.

This is a pretty big change, sadly I don't have the time to check it out on my PC so I can't add much to this discussion. Looks like a step in the right direction though :+1:

markusressel (approved):

natario1:
Seems good and much cleaner. Will add rotation at some point thanks to this

markusressel:
@natario1 btw it would be nice if you could wait with deleting PR branches until CI is done.
Otherwise it looks like you merged a failing PR which is quite irritating.

Issue #101, opened 2019-03-29
Allow pan with only two or more fingers


RayyanTahir:
I'm writing a screen sharing/viewing app like teamviewer where I can control the remote stream, I have the mouse movement set up with single finger pan, but with this library attached the image pans as well. Is there a way to only allow panning when two or more fingers are used?

Any kind of help will be appreciated. Thanks

natario1:
There is no option for this at the moment, sorry! If you want to work on it, it would be cool to have two attributes, `scrollEnabled` (generic but currently missing) and `singleFingerScrollEnabled` (which you could then disable).

RayyanTahir:
Thank you, I could give it a shot and come up with a PR maybe?

natario1:
That would be cool! Thank you!

PR #102, opened 2019-04-01
Added scrollEnabled, oneFingerScrollEnabled, twoFingersScrollEnabled and threeFingersScrollEnabled

RayyanTahir:
This PR adds:

1. `scrollEnabled` if set to false, scrolling will be disabled. This is set to true by default
2. `oneFingerScrollEnabled` if set to false, scrolling cannot be done with one finger. This is set to true by default
3. `twoFingersScrollEnabled` if set to false, scrolling cannot be done with two fingers. This is set to true by default
4. `threeFingersScrollEnabled` if set to false, scrolling cannot be done with three fingers. This is set to true by default

Fixes/Satisfies feature request:
#101


RayyanTahir:
@natario1 Please review this when you have time. Thanks

markusressel:
Isn't the naming convention a bit confusing when we already have `horizontalPanEnabled` and `horizontalPanEnabled`?

markusressel (changes requested):
The code looks fine to me, I only have an issue with the naming of those attributes.
Since @natario1 most likely want's to have a stable API we can't rename existing attributes so I would suggest to call yours `...Pan...` instead of `...Scroll...`.

And: those attributes have to be added to the documentation too.

library/src/main/java/com/otaliastudios/zoom/ZoomApi.kt
```diff
@@ -229,6 +229,20 @@ interface ZoomApi {
     fun setFlingEnabled(enabled: Boolean)
     // TODO (v2) rename to var isFlingEnabled or isGestureFlingEnabled to clearly differentiate gestures and engine movements (pan/zoom)
 
+    /**
+     * Controls whether scroll gesture is enabled or not.
+     *
+     * @param enabled true enables scroll gesture, false disables it
+     */
+    fun setScrollEnabled(enabled: Boolean)
```
natario1, line 237: Can you add these two in README?

library/src/main/java/com/otaliastudios/zoom/internal/gestures/ScrollFlingDetector.kt
```diff
@@ -144,6 +146,8 @@ internal class ScrollFlingDetector(
             @ZoomApi.ScaledPan distanceX: Float,
             @ZoomApi.ScaledPan distanceY: Float
     ): Boolean {
+        if (!scrollEnabled) return false
+        if (!singleFingerScrollEnabled && e1 != null && e1.pointerCount == 1) return false
```
natario1: You should use e2 here, as e1 is the starting event and should always have 1 pointer.

Could you write it like this to make it more readable:

val isSingleFinger = e2?.pointerCount == 1
if (!singleFingerScrollEnabled && isSingleFinger) return false

natario1 (changes requested):
I think naming is OK! We want to have both engine movement control (zoomEnabled, panEnabled) and selective control over the gestures that create these movements (flingEnabled, scrollEnabled, ... ).

Now that I see this, it would be cool to also have `doubleFingerScrollEnabled` and `tripleFingerScrollEnabled`, if you have time. If not, please just address the comments above. Thanks!

library/src/main/java/com/otaliastudios/zoom/internal/gestures/ScrollFlingDetector.kt
RayyanTahir: Sure, I'll fix it

library/src/main/java/com/otaliastudios/zoom/ZoomApi.kt
RayyanTahir: I'll add it to the README

RayyanTahir:
Thankyou for your feedbacks @natario1 @markusressel. If the naming convention is alright, I'll move forward with double and triple finger scroll properties along with the changes required in README and motion event handling

natario1:
Maybe oneFingerScroll, twoFingersScroll and threeFingersScroll are better than single, double and triple. I would do this way

natario1:
I just read your first comment @markusressel I had missed it.

I don't see this as redundant, pan is an engine movement that can be generated by a few gestures - flinging, single finger scroll, two fingers scroll, three fingers scroll, and even while pinching. We're giving selective control here.

For instance, we also already have flingEnabled. In a future release we can use the prefix "gesture" for all these APIs (or "movement" suffix for pan and zoom instead, like panMovementEnabled), but for now I think docs will be enough, what do you think?

This is reflected in code, as these flags go to gesture detectors, while movement flags go to movement managers.

markusressel:
@natario1 sounds good to me :+1:

natario1:
Cool! When this is merged I will add the missing flags, which are scrollWhilePinchingEnabled and pinchEnabled. pinchEnabled is quite redundant with zoomEnabled at this point but maybe in the future we can implement other ways to zoom. For example google maps lets you zoom with a single finger (tap + second tap and hold)

So @RayyanTahir , you can go on!

RayyanTahir:
Thankyou, @natario1 @markusressel please review my changes. 

1. I have added details of flags to README
2. Fixed motion event handling to detect pointer count using e2
3. Used the new naming convention; oneFingerScrollEnabled, twoFingersScrollEnabled and threeFingersScrollEnabled

library/src/main/java/com/otaliastudios/zoom/ZoomApi.kt
```diff
@@ -237,11 +237,25 @@ interface ZoomApi {
...
-     * @param enabled true enables single finger scroll gesture, false disables it
+     * @param enabled true enables one finger scroll gesture, false disables it
      */
-    fun setSingleFingerScrollEnabled(enabled: Boolean)
+    fun setOneFingerScrollEnabled(enabled: Boolean)
+
+    /**
+     * Controls whether two fingers scroll gesture is enabled or not.
+     *
+     * @param enabled true enables twi fingers scroll gesture, false disables it
```
markusressel: typo: **two**

library/src/main/res/values/attrs.xml
```diff
@@ -10,7 +10,9 @@
         <attr name="zoomEnabled" format="boolean" />
         <attr name="flingEnabled" format="boolean" />
         <attr name="scrollEnabled" format="boolean" />
-        <attr name="singleFingerScrollEnabled" format="boolean" />
+        <attr name="oneFingerScrollEnabled" format="boolean" />
+        <attr name="twoFingersScrollEnabled" format="boolean" />
```
markusressel, line 14: do you really want to go with finger vs finger**s**?
`twoFingerScrollEnabled` would be ok in my brain and it would be more consistent only changing the count.

library/src/main/java/com/otaliastudios/zoom/ZoomApi.kt
RayyanTahir: 🤦‍♂️ my bad. Fixing

library/src/main/res/values/attrs.xml
RayyanTahir: While it would make sense to go with `twoFingerScrollEnabled`, I think we should still go with `twoFingersScrollEnabled` since @natario1 specifically suggested here https://github.com/natario1/ZoomLayout/pull/102#issuecomment-478974434

RayyanTahir:
@natario1 @markusressel Is everything good to go for merge?

markusressel:
I want to wait for @natario1 opinion on the naming subject because if we want to change this later it would be a breaking change. Other than that its fine.

natario1 (approved):

natario1:
It's fine for me! Name seems OK... Thanks for your work @RayyanTahir !

PR #157, opened 2020-06-12
raise an exception if a drawable without intrinsic dimensions is used

markusressel:
Since drawables without intrinsic dimensions can not be used in the `ZoomImageView` I think it is better to raise an exception if a user tries to do just that, instead of silently "failing".


natario1 (approved):
Agree

Issue #47, opened 2018-07-24
Need protip: how to panTo zoomed view to specific point? (10% for example)


Svoka:
Hi! Can't figure it out which params I need to use to calculate panTo values?

Size of the screen and realSize of the view?

natario1:
Just the size of your content, in pixels.

If using ZoomLayout, it's the layout_width and layout_height of the view that you have put inside it. 
If using ZoomImageView, it's the drawable size.

Note that a positive translation will move the content down/right. So depending on your case, you might need to use negative values.

Svoka:
Hm, 
here is my code

` panX = -(l_content.width  / 2).toFloat()

 panY = -(l_content.height  / 2).toFloat()

zm_zoom.engine!!.moveTo(zoom, panX, panY, false)`

I'm expect content to be zoom on the center of the screen, but it's just on the right/bottom position,
do I miss something?

natario1:
This will move the center of the content to the top-left corner of ZoomLayout
(assuming it is possible at the given zoom level. We never allow empty space on more than 2 sides)

natario1:
Something like this might work.

```kotlin
val zoomLayoutCenterX = (zoomLayout.width / 2).toFloat()
val zoomLayoutCenterY = (zoomLayout.height / 2).toFloat()
val contentCenterX = (content.width / 2).toFloat()
val contentCenterY = (content.height / 2).toFloat()
val diffX = contentCenterX - zoomLayoutCenterX
val diffY = contentCenterY - zoomLayoutCenterY
zoomLayout.panTo(-diffX, -diffY, true)
```

Anyway everything is possible, it just takes some logic.

Svoka:
@natario1 thanks! I knew i need the part with screen size (zoomLayout.width in your example)

PR #158, opened 2020-06-12
Bugfix/#152 pan to center

### LINKED ISSUE #152: ZoomImageView doesn't initially pan to center of the bitmap
- ZoomLayout 1.7.0/1
- Reproducible in official demo app: *Yes*
- Device / Android version: *Pixel, Android 10*

> For some reason when setting the image on the ZoomImageView, it doesn't pan to the center. It'll pan to the top, or bottom, but never the center regardless of what gravity or alignment I set.

> Currently the only way I can get it to center when initially setting the bitmap is to first call zoomImageView.setImageBitmap(bitmap) then set the pan to zoomImageView.panX / 2 and zoomImageView.panY / 2. This seems to work when the width and height of the content are bigger than the container(zoomImageView), however when either the width or height of the content is smaller than the container, the panx and pany both end up being 0, so this solution doesn't work. 

I recently opened a GitHub issue that ended up being closed as stale https://github.com/natario1/ZoomLayout/issues/135. Basically, I've gotten back to trying to repro the issue and now have a version of the sample app in which you'll be able to repro the issue.

https://github.com/King-Henry/ZoomLayout-Centering-Issue. This is the repository that contains the code to reproduce.

markusressel:
Honestly, I still can't quite wrap my head around how the matrix transformation stuff works, so I am not entirely sure if this is the right thing to do to fix the issue.

I tested the `ZoomImageView` on my device with different types of images of different size, and it works for all of them. I have **not** tested if code based functions like `panTo` or `moveTo` are affected by this. If I should do so please let me know.

It would probably be a good idea to add tests for these transformation functions since - at least for outside contributors - it is probably really hard to understand their intended and correct behavior, and a tiny change in code can have a massive impact on the functionality.

- Fixes #152
- Tests: *no*
- Docs updated: *no*

### Solution
As I discovered that your proposal code for centering an image in #47 did not work I found the code in `computeTransformationPan()` that did the exact same thing. Applying the `realZoom` to the content to match its `width` and `height` values with the `width` and `height` of the container fixed the issue.

Again, I don't understand every line of code in this lib yet, so, please have a thorough look and let me know if anything about this change worries you.

### DIFF
--- a/library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
+++ b/library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
@@ -76,13 +76,9 @@ internal constructor(context: Context) : ZoomApi {
 
         // Post utilities
 
-        override fun post(action: Runnable) {
-            container.post(action)
-        }
+        override fun post(action: Runnable): Boolean = container.post(action)
 
-        override fun postOnAnimation(action: Runnable) {
-            container.postOnAnimation(action)
-        }
+        override fun postOnAnimation(action: Runnable) = container.postOnAnimation(action)
 
         // Matrix callbacks
 
--- a/library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
+++ b/library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
@@ -4,13 +4,16 @@ import android.annotation.SuppressLint
 import android.content.Context
 import android.graphics.Matrix
 import android.graphics.RectF
-import android.view.*
+import android.view.Gravity
+import android.view.MotionEvent
+import android.view.View
+import android.view.ViewTreeObserver
 import com.otaliastudios.zoom.ZoomApi.*
-import com.otaliastudios.zoom.internal.UpdatesDispatcher
-import com.otaliastudios.zoom.internal.matrix.MatrixController
 import com.otaliastudios.zoom.internal.StateController
+import com.otaliastudios.zoom.internal.UpdatesDispatcher
 import com.otaliastudios.zoom.internal.gestures.PinchDetector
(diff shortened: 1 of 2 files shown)


markusressel:
Thinking about this, could this also be an issue in `computeTransformationZoom()`?

library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt
```diff
@@ -607,8 +606,8 @@ internal constructor(context: Context) : ZoomApi {
      * dimensions. This means applying the transformation gravity.
      */
     private fun computeTransformationPan(): ScaledPoint {
-        val extraWidth = contentWidth - containerWidth
-        val extraHeight = contentHeight - containerHeight
+        val extraWidth = contentWidth * realZoom - containerWidth
+        val extraHeight = contentHeight * realZoom - containerHeight
```
markusressel, line 610: These are the important 2 lines.

natario1 (approved):
If it helps, anytime I go back to this repo after some time, I don't understand the code either. And that's because of bad naming I think (like, what the hell is a realZoom?). We should go through a big change and rethink/rename APIs at some point.

About this PR I'm a bit confused as to why it seemed to work just fine until now (the demo app correctly centers the image IIRC). But the changes you made seem to make sense to me, so if you know they fix the issue, let's merge this!

markusressel:
My guess is that the issue is related to the dimensions of the device, the image, or the fact that multiple images are applied in succession (setImageDrawable) or a combination of those things.

natario1:
I see. Feel free to merge when you're ready

markusressel:
I have done some additional monkey testing and haven't found any issues, so I will merge and we will see if something explodes :sweat_smile:

PR #164, opened 2020-06-25
Feature/#100 overzoom customization

### LINKED ISSUE #100: DEFAULT_OVERZOOM_FACTOR customizable
Hi!
Is there any way to set up the `DEFAULT_OVERZOOM_FACTOR` in order to increase the zoom animation when `overPinchable`is true?

markusressel:
- Fixes #100
- Tests: *no*
- Docs updated: *yes*

### Solution
The overscroll and overzoom amount can now be specified using a `OverPanRangeProvider` and `OverZoomRangeProvider` respectively.

While the `OverZoomRangeProvider` has only one method that defines the range in both directions, the `OverPanRangeProvider` has two methods, one for vertical and one for the horizontal range. I think it would be a good idea to allow this separation for the zoom-in and zoom-out.

When implementing and testing, I realized that with the current implementation it would not be possible for a developer to recreate the behavior of the `DEFAULT_OVERZOOM_PROVIDER`, since there is no way for a dev to access the current `maxZoom` and `minZoom` values. It is only possible to set these, but they can not be queried, since the methods are `internal`. I am not sure how you would want to deal with this, if at all. Since these values also depend on the `ZoomType` it might be difficult to expose a single method for this.
EDIT:
I just realized you mentioned this in a comment, stating that the `maxZoom` and `minZoom` values would need public getters. I am still not sure how to deal with the zoom type though.

I am also a little bit torn about the naming. In the docs, things are called "Scroll", but in the code they are often called "Pan" instead and I am not sure if they are supposed to be the same thing or not.

### DIFF
--- a/library/src/main/java/com/otaliastudios/zoom/OverPanRangeProvider.kt
+++ b/library/src/main/java/com/otaliastudios/zoom/OverPanRangeProvider.kt
@@ -1,0 +1,20 @@
+package com.otaliastudios.zoom
+
+/**
+ * Defines the allowed range for overpan.
+ */
+interface OverPanRangeProvider {
+
+    /**
+     * @return the maximum horizontal overpan to allow
+     */
+    @ZoomApi.ScaledPan
+    fun getHorizontalOverPanRange(engine: ZoomEngine): Float
+
+    /**
+     * @return the maximum vertical overpan to allow
+     */
+    @ZoomApi.ScaledPan
+    fun getVerticalOverPanRange(engine: ZoomEngine): Float
+
+}
--- a/library/src/main/java/com/otaliastudios/zoom/OverPanRangeProvider.kt
+++ b/library/src/main/java/com/otaliastudios/zoom/OverPanRangeProvider.kt
@@ -6,15 +6,29 @@ package com.otaliastudios.zoom
 interface OverPanRangeProvider {
 
     /**
-     * @return the maximum horizontal overpan to allow
+     * Calculates the maximum overpan range
+     *
+     * @param engine the zoom engine
+     * @param horizontal true when horizontal range should be calculated, false for vertical
+     * @return the maximum overpan to allow
      */
     @ZoomApi.ScaledPan
-    fun getHorizontalOverPanRange(engine: ZoomEngine): Float
+    fun getOverPanRange(engine: ZoomEngine, horizontal: Boolean): Float
 
-    /**
-     * @return the maximum vertical overpan to allow
-     */
-    @ZoomApi.ScaledPan
-    fun getVerticalOverPanRange(engine: ZoomEngine): Float
+    companion object {
+        // TODO add OverScrollCallback and OverPinchCallback.
+        // Should notify the user when the boundaries are reached.
+        // TODO expose friction parameters, use an interpolator.
+        // TODO Make public, add API.
+        /**
+         * The default overscrolling factor
+         */
+        private const val DEFAULT_OVERPAN_FACTOR = 0.10f
+        val DEFAULT = object : OverPanRangeProvider {
+            override fun getOverPanRange(engine: ZoomEngine, horizontal: Boolean): Float {
+                return engine.containerHeight * DEFAULT_OVERPAN_FACTOR
+            }
+        }
+    }
 
 }
--- a/library/src/main/java/com/otaliastudios/zoom/OverPanRangeProvider.kt
+++ b/library/src/main/java/com/otaliastudios/zoom/OverPanRangeProvider.kt
@@ -6,29 +6,30 @@ package com.otaliastudios.zoom
 interface OverPanRangeProvider {
 
     /**
-     * Calculates the maximum overpan range
+     * Calculates the maximum overpan
      *
      * @param engine the zoom engine
      * @param horizontal true when horizontal range should be calculated, false for vertical
      * @return the maximum overpan to allow
      */
     @ZoomApi.ScaledPan
-    fun getOverPanRange(engine: ZoomEngine, horizontal: Boolean): Float
+    fun getOverPan(engine: ZoomEngine, horizontal: Boolean): Float
 
     companion object {
         // TODO add OverScrollCallback and OverPinchCallback.
         // Should notify the user when the boundaries are reached.
         // TODO expose friction parameters, use an interpolator.
         // TODO Make public, add API.
-        /**
-         * The default overscrolling factor
-         */
-        private const val DEFAULT_OVERPAN_FACTOR = 0.10f
+
+        @JvmField
         val DEFAULT = object : OverPanRangeProvider {
-            override fun getOverPanRange(engine: ZoomEngine, horizontal: Boolean): Float {
-                return engine.containerHeight * DEFAULT_OVERPAN_FACTOR
+            private val DEFAULT_OVERPAN_FACTOR = 0.10f
+            override fun getOverPan(engine: ZoomEngine, horizontal: Boolean): Float {
+                return when (horizontal) {
+                    true -> engine.containerWidth * DEFAULT_OVERPAN_FACTOR
+                    false -> engine.containerHeight * DEFAULT_OVERPAN_FACTOR
+                }
             }
         }
     }
-
 }
--- a/library/src/main/java/com/otaliastudios/zoom/OverZoomRangeProvider.kt
+++ b/library/src/main/java/com/otaliastudios/zoom/OverZoomRangeProvider.kt
@@ -16,8 +16,9 @@ interface OverZoomRangeProvider {
     fun getOverZoom(engine: ZoomEngine, inwards: Boolean): Float
 
     companion object {
-        private const val DEFAULT_OVERZOOM_FACTOR = 0.1f
+        @JvmField
         val DEFAULT = object : OverZoomRangeProvider {
+            private val DEFAULT_OVERZOOM_FACTOR = 0.1f
             override fun getOverZoom(engine: ZoomEngine, inwards: Boolean): Float {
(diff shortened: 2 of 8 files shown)


natario1:
This is great! 

For zoom getters, I think you can just expose them. In the future (like in a v2) I would like to:
- change REAL_ZOOM to be just ZOOM. Since it's the actual zoom, let's call it zoom!
- change ZOOM to be RELATIVE_ZOOM since it's relative to the transformation

So the current `minZoom` and `maxZoom` would be consistent with our future plans at least 😄 

For pan vs. scroll, I like `pan` better. I think we should distinguish between the movement **type** (pan, zoom, maybe rotation in the future) and what can cause it (scroll and fling can both cause pan, pinch can cause zoom/rotation). This provider you're adding is related to the movement type so pan is fine IMO

markusressel:
@natario1 I have exposed `minZoom`, `minZoomType`, `maxZoom` and `maxZoomType` via `ZoomApi` now.
Also I changed the `OverZoomRangeProvider` to contain `getOverZoomIn` and `getOverZoomOut` to provide separate functions for inwards and outwards zoom.
Please have a look :)

library/src/main/java/com/otaliastudios/zoom/OverPanRangeProvider.kt
```diff
@@ -0,0 +1,20 @@
...
+     * @return the maximum horizontal overpan to allow
+     */
+    @ZoomApi.ScaledPan
+    fun getHorizontalOverPanRange(engine: ZoomEngine): Float
+
+    /**
+     * @return the maximum vertical overpan to allow
+     */
+    @ZoomApi.ScaledPan
+    fun getVerticalOverPanRange(engine: ZoomEngine): Float
```
natario1: I would probably prefer if we had just one function with a boolean parameter. More handy for people who want to use the same value for horizontal/vertical (the majority I think).

library/src/main/java/com/otaliastudios/zoom/OverZoomRangeProvider.kt
```diff
@@ -0,0 +1,20 @@
...
+     * @return the maximum inwards overzoom to allow
+     */
+    @ZoomApi.RealZoom
+    fun getOverZoomIn(engine: ZoomEngine): Float
+
+    /**
+     * @return the maximum outwards overzoom to allow
+     */
+    @ZoomApi.RealZoom
+    fun getOverZoomOut(engine: ZoomEngine): Float
```
natario1: Same here about one function + boolean

library/src/main/java/com/otaliastudios/zoom/internal/movement/PanManager.kt
```diff
@@ -185,5 +196,15 @@ internal class PanManager(provider: () -> MatrixController) : MovementManager(pr
...
+
+        val DEFAULT_OVERPAN_PROVIDER = object : OverPanRangeProvider {
+            override fun getHorizontalOverPanRange(engine: ZoomEngine): Float {
+                return engine.containerWidth * DEFAULT_OVERPAN_FACTOR
+            }
+
+            override fun getVerticalOverPanRange(engine: ZoomEngine): Float {
+                return engine.containerHeight * DEFAULT_OVERPAN_FACTOR
+            }
+        }
```
natario1: Can we move this to OverPanRangeProvider interface so people can call `OverPanRangeProvider.DEFAULT`?

library/src/main/java/com/otaliastudios/zoom/internal/movement/ZoomManager.kt
```diff
@@ -115,13 +129,23 @@ internal class ZoomManager(provider: () -> MatrixController) : MovementManager(p
...
+        const val DEFAULT_OVERZOOM_FACTOR = 0.1f
+        val DEFAULT_OVERZOOM_PROVIDER = object : OverZoomRangeProvider {
+            override fun getOverZoomIn(engine: ZoomEngine): Float {
+                return DEFAULT_OVERZOOM_FACTOR * (engine.getMaxZoom() - engine.getMinZoom())
+            }
+
+            override fun getOverZoomOut(engine: ZoomEngine): Float {
+                return DEFAULT_OVERZOOM_FACTOR * (engine.getMaxZoom() - engine.getMinZoom())
+            }
+        }
```
natario1: Can we move this to OverZoomRangeProvider interface so people can call `OverZoomRangeProvider.DEFAULT`? Also I don't think the factor should be public (it's not like a carefully tuned value) but if you need it public, let's make the pan factor public too.

library/src/main/java/com/otaliastudios/zoom/internal/movement/ZoomManager.kt
```diff
@@ -72,12 +78,20 @@ internal class ZoomManager(provider: () -> MatrixController) : MovementManager(p
...
+
+    /**
+     * The amount of overzoom that is allowed in outwards direction.
+     * This value is calculated by the [overZoomRangeProvider].
      */
     @ZoomApi.RealZoom
-    internal val maxOverZoom: Float
-        get() = DEFAULT_OVERZOOM_FACTOR * (getMaxZoom() - getMinZoom())
+    internal val maxOverZoomOut: Float
+        get() = overZoomRangeProvider.getOverZoomOut(engine)
```
natario1: I think we should use abs() here to ensure its >= 0

library/src/main/java/com/otaliastudios/zoom/internal/movement/PanManager.kt
```diff
@@ -144,16 +153,18 @@ internal class PanManager(provider: () -> MatrixController) : MovementManager(pr
...
      */
     @ZoomApi.ScaledPan
-    internal val maxOverPan: Float
-        get() {
-            val overX = controller.containerWidth * DEFAULT_OVERPAN_FACTOR
-            val overY = controller.containerHeight * DEFAULT_OVERPAN_FACTOR
-            return min(overX, overY)
-        }
+    internal val maxHorizontalOverPan: Float
+        get() = overPanRangeProvider.getHorizontalOverPanRange(engine)
```
natario1: Same here, I think we should abs() the user value. Or max(0, value) ?

natario1 (changes requested):

library/src/main/java/com/otaliastudios/zoom/internal/movement/PanManager.kt
markusressel: I like max(0, value) better. I will also add a warning log.

library/src/main/java/com/otaliastudios/zoom/OverPanRangeProvider.kt
```diff
@@ -0,0 +1,34 @@
...
+        // Should notify the user when the boundaries are reached.
+        // TODO expose friction parameters, use an interpolator.
+        // TODO Make public, add API.
+        /**
+         * The default overscrolling factor
+         */
+        private const val DEFAULT_OVERPAN_FACTOR = 0.10f
+        val DEFAULT = object : OverPanRangeProvider {
+            override fun getOverPanRange(engine: ZoomEngine, horizontal: Boolean): Float {
+                return engine.containerHeight * DEFAULT_OVERPAN_FACTOR
```
natario1: Should we use containerWidth when horizontal?

library/src/main/java/com/otaliastudios/zoom/OverPanRangeProvider.kt
```diff
@@ -0,0 +1,34 @@
...
+
+    /**
+     * Calculates the maximum overpan range
+     *
+     * @param engine the zoom engine
+     * @param horizontal true when horizontal range should be calculated, false for vertical
+     * @return the maximum overpan to allow
+     */
+    @ZoomApi.ScaledPan
+    fun getOverPanRange(engine: ZoomEngine, horizontal: Boolean): Float
```
natario1: I'd remove "Range" from the method name, for consistency with the zoom one.

library/src/main/java/com/otaliastudios/zoom/OverZoomRangeProvider.kt
```diff
@@ -0,0 +1,27 @@
...
+     * @param engine the zoom engine
+     * @param inwards true for inwards, false for outwards
+     * @return the maximum overzoom to allow
+     */
+    @ZoomApi.RealZoom
+    fun getOverZoom(engine: ZoomEngine, inwards: Boolean): Float
+
+    companion object {
+        private const val DEFAULT_OVERZOOM_FACTOR = 0.1f
+        val DEFAULT = object : OverZoomRangeProvider {
```
natario1, line 20: I wonder if we need some Jvm annotation to make this accessible from java as a static field?

natario1 (approved):
I left a couple of questions but overall looks good to me, feel free to merge when you're ready

library/src/main/java/com/otaliastudios/zoom/OverPanRangeProvider.kt
markusressel: good catch, got lost in copy pasting :S

library/src/main/java/com/otaliastudios/zoom/OverZoomRangeProvider.kt
markusressel: Hmm, yes indeed it does not work by default.
I tried to use `@JvmField`

but it states:
```
JvmField could be applied only if all interface companion properties are 'public final val' with '@JvmField' annotation
```
which is kinda odd imo. To solve this I moved the private constant into the definition of the default `OverZoomRangeProvider`.

markusressel:
Although I don't understand why your approval is still valid, even though I changed the code, I will go ahead and merge this since you stated that you are fine with it. If something is wrong, please let me know.

Issue #165, opened 2020-06-25
Back Image to the center after movements


Yahor10:
### How do I?

I set aligment to none and after some MOVE Actions want toback my image to center (origin position).

Image does not move back  with code `zoomLayout.setAlignment(Alignment.CENTER);
`to the center.

How to do it correct?
### Version used
Last Version

natario1:
We don't have an API to re-apply the transformation by you can try re-setting the size: `zoomLayout.engine.setContainerSize(zoomLayout.width, zoomLayout.height, true)`. Note the true at the end.

markusressel:
@natario1 what do you think about adding a helper function to the to the api for this use case, like f.ex. `moveToCenter(zoom: Float? = null)`? It seems like a hard thing to do for newcomers, probably because of the way `moveTo` uses the left upper corner as a reference, instead of the center of the view.

Yahor10:
`zoomLayout.engine.setContainerSize(zoomLayout.width, zoomLayout.height, true)`
Works better. Effect is sharp still.

How to do it with smooth animation?

natario1:
@Yahor10 you have to use the move APIs to zoom where you want.
@markusressel please do if you want. We could also maintain a docs page with "recipes", since we can't cover all cases with APIs and we should really encourage people to do calculations and go through a few minutes of thinking, concepts are simple and understanding them will unlock all the possibilities.

markusressel:
Closing this since everything has been said.

PR #167, opened 2020-06-27
Feature/#165 move to center

markusressel:
This adds a `moveToCenter` api method, as described in #165.

Btw, I have thought about your "recipes" idea, and - while I fully support this idea - I think, better than only writing a tutorial, is to write code in `ZoomLayout`, that actually implements these ideas, create a write-up of what the initial goal was, what the problems are and explain how the code works with a link to the actual `ZoomLayout` implementation (I mean the library here, not the class). That way the written code is always up to date, tested by actual `ZoomLayout` users, and migrated along the way as the library progresses. That way it can be used both as a functional API interface by users, as well as the code for a tutorial. Of course it has to be evaluated if a problem is worth including in the library, which is a fine.


natario1 (approved):
Looks good!

Issue #188, opened 2020-10-21
moveTo()'s animation looks a little weird


ln206:
### How do I?
When I used moveTo(), the animation doesn't look linear. It looks like it was swirling for a bit. How do I get rid of this swirl effect? Is there a solution for this?

### Version used
1.8.0

markusressel:
Well not without a PR or a custom implementation on your end :smile: 
I think this is caused by the interaction of concurrent pan and zoom animations, and there is probably a solution that, but we haven't been able to tackle that yet. I'll leave this ticket open since I have noticed this myself for a long time now.

markusressel:
This was already mentioned in #113, but went stale over time. I think it is a good idea to keep this open.

markusressel:
I did a little digging.

Turns out the animation duration of the pan and zoom don't even align. When using some really long animation durations it becomes obvious that the pan animation finishes much quicker than the zoom animation, although the `animatedFraction` value of the animator is not all the way to `1` yet.

I did some testing, and - after almost losing all of my hair - I found that the `PropertyValuesHolder.ofObject` method (in `internal fun animateUpdate(update: MatrixUpdate)`) seems to do weird stuff, as it doesn't really behave like expected. Even when forcing a `LinearInterpolator`, the animation of the pan does not look linear (still looks kina accellerated), while the zoom does. I tried exchanging the pan property holder for two `PropertyValuesHolder.ofFloat` holders, and voila, the movement is indeed linear, and it also takes up the full animation duration, as expected.

Maybe I am not seeing something really obvious here and we just use the `ofObject` method wrong, but I would just ditch the `ofObject` method for now. Although even when applying this the `moveTo` animation is still a little weird at times, it doesn't do the strange wiggle anymore. The biggest issue I see is that it sometimes "bashes against the walls", which could probably be fixed by aligning the input coordinates with the boundaries of the current content size, or even simpler by allowing overpan temporarily during this animation. Also, the user defined animation duration is actually respected for both pan and zoom.

Yahor10:
To be clear:
When i use image 512x512 and wanna zoom to Point (200,200);
First i `zoomLayout.zoomTo(2.0);`
Second` zoomLayout.panTo(-200),-200);
Or 
I should use 

`zoomLayout.panTo(-200 *2.0,-200*2.0);    //2.0 becaouse of scale ?

PR #189, opened 2020-11-04
Less jiggly animation on moveTo

markusressel:
- Contributes to #188
- Tests: *no*
- Docs updated: *no*

### Solution

This replaces the usage of `PropertyValuesHolder.ofObject` with two `PropertyValuesHolder.ofFloat` instances, to work around the weird behavior which is not respecting the set animation interpolator.

This also pretty much eliminates the "wiggly" animation when using `moveTo`, although it still has problems, like "bashing" into walls due to overpan not being allowed during the animation.

### DIFF
--- a/library/src/main/java/com/otaliastudios/zoom/internal/matrix/MatrixController.kt
+++ b/library/src/main/java/com/otaliastudios/zoom/internal/matrix/MatrixController.kt
@@ -322,10 +322,12 @@ internal class MatrixController(
         val holders = mutableListOf<PropertyValuesHolder>()
         if (update.pan != null) {
             val target = if (update.isPanRelative) pan + update.pan else update.pan
+            // ofObject doesn't respect animator.interpolator, so we use ofFloat instead
             holders.add(PropertyValuesHolder.ofFloat("panX", panX, target.x))
             holders.add(PropertyValuesHolder.ofFloat("panY", panY, target.y))
         } else if (update.scaledPan != null) {
             val target = if (update.isPanRelative) scaledPan + update.scaledPan else update.scaledPan
+            // ofObject doesn't respect animator.interpolator, so we use ofFloat instead
             holders.add(PropertyValuesHolder.ofFloat("panX", scaledPanX, target.x))
             holders.add(PropertyValuesHolder.ofFloat("panY", scaledPanY, target.y))
         }
--- a/library/src/main/java/com/otaliastudios/zoom/internal/matrix/MatrixController.kt
+++ b/library/src/main/java/com/otaliastudios/zoom/internal/matrix/MatrixController.kt
@@ -302,14 +302,6 @@ internal class MatrixController(
         activeAnimators.clear()
     }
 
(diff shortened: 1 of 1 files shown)


markusressel:
Maybe it would be a good idea to add a comment about this to the code, since I will certainly forget about this detail in T-10 seconds...

natario1 (approved):
Comment would help! No idea about why this happens.

On a side note, should AbsolutePoint and ScaledPoint be inline classes? I wonder if this would avoid JVM allocations which are especially painful during animations.

markusressel:
> On a side note, should AbsolutePoint and ScaledPoint be inline classes? I wonder if this would avoid JVM allocations which are especially painful during animations.

Probably, however, [according to the Kotlin docs](https://kotlinlang.org/docs/reference/inline-classes.html):
> An inline class must have a single property initialized in the primary constructor.

Which doesn't seem like a very intuitive thing for an object that is meant to describe two values.

natario1:
Sounds good 😄

luisangelsm:
Hey, I have been testing `ZoomLayout` and I don't think this problem is actually fixed. If you use the sample app and call `zoomLayou.getEngine().moveToCenter` you can clearly see how the pan animation is completely wrong, the center of the content moves around during the animation.

There are also hiccups when the content hits the edge of the container, I think they are what @markusressel calls `bashing into walls`. Enabling `overpan` doesn't really fix these jumps (and the animation can easily end in a wrong state), after logging the actual pan applied to the view you can see some weirdness in it.

@natario1 @markusressel Any ideas about what is going on? I have been trying to fix the problem(s), any help is really welcome.

markusressel:
Well, since it was over 2 years ago when I last looked at this, sadly I have no idea :cry: All of the research that I have done has been documented in #188, and it may very well not be trivial to fix this this.

luisangelsm:
Thanks for the reply @markusressel 

One thing I think it causes problems is trying to fix zoom and pan during the animation (`checkBounds` and `ensurePan`) and maybe other calculations inside the pan manager. Before starting the animation the library should calculate a valid end state in advance. Not all zoom and pan combinations can be fulfilled, but given a target zoom and a target pan it should be possible to know exactly the final state, and the animation should happen between those two states allowing any intermediate state even if it is not valid. I am not sure if this is something it can just happen inside `MatrixController` or if the zoom/pan managers need tweaking.

This is what happens with a zoom + pan animation, it is the actual panX that the ZoomLayout is getting. It is full of bumps before reaching 0, where you can clearly see a hard transition, an then you get a nice curve from there. I think those jumps happen because of the corrections applied during the animation.

x

That hard transition is probably caused by how the gravity affects the position of the content view before it gets bigger than the container view, gravity should not have any effect during the animation and it just needs to be considered when calculating the final state before hand.

chengxuyuanluyu:
so,PropertyValuesHolder.ofObject

chengxuyuanluyu:
PropertyValuesHolder.ofObject

chengxuyuanluyu:
Is there a good solution now

PR #214, opened 2021-11-24
Fixed verbose logging via ZoomLogger

r4zzz4k:
Currently verbose logs are sent to `Log.i`, but the message validation is still based on verbose level. So:
* With `ZoomLogger.setLogLevel(ZoomLogger.LEVEL_VERBOSE)`, all the messages are there, though they are posted as info level, so it's a bit hard to make sense of my own debug logging interleaved with these.
* With `ZoomLogger.setLogLevel(ZoomLogger.LEVEL_INFO)`, there are tons of empty `I/StateController: ` in logcat.

This PR is a simple fix for the issue, calling correct `Log` method for verbose logs.
No need for tests and docs updates.


natario1 (approved):

natario1:
Thanks! 👍

PR #227, opened 2022-07-26
设置缩放级别时，添加参数用来控制是否需要立刻缩放控件

konstant2016:
我们的项目中遇到的需求，可以动态修改最大缩放级别，但是在当前源码中，修改缩放级别后，内部view会立刻执行缩放操作，因此加上参数后可以用来控制是否立刻进行缩放

### DIFF
--- a/library/src/main/java/com/otaliastudios/zoom/ZoomApi.kt
+++ b/library/src/main/java/com/otaliastudios/zoom/ZoomApi.kt
@@ -418,7 +418,7 @@ interface ZoomApi {
      *
      * @param maxZoom the max zoom
      */
-    fun setMaxZoom(@Zoom maxZoom: Float) {
+    fun setMaxZoom(@Zoom maxZoom: Float,scaleView:Boolean = true) {
         setMaxZoom(maxZoom, TYPE_ZOOM)
     }
 
@@ -432,7 +432,7 @@ interface ZoomApi {
      * @see zoom
      * @see realZoom
      */
-    fun setMaxZoom(maxZoom: Float, @ZoomType type: Int)
+    fun setMaxZoom(maxZoom: Float, @ZoomType type: Int,scaleView:Boolean = true)
 
     /**
      * Get the currently allowed min zoom.
@@ -459,7 +459,7 @@ interface ZoomApi {
      *
      * @param minZoom the min zoom
      */
-    fun setMinZoom(@Zoom minZoom: Float) {
+    fun setMinZoom(@Zoom minZoom: Float,scaleView:Boolean = true) {
         setMinZoom(minZoom, TYPE_ZOOM)
     }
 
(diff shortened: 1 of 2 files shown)


markusressel:
Automatic translation:

> In our project, we have a requirement to dynamically modify the maximum zoom level, but in the current source code, the internal view will perform the zoom operation immediately after the zoom level is modified, so adding the parameter can be used to control whether the zoom is performed immediately

konstant2016:
已经成功收到您的邮件

markusressel:
The reason the scale is performed is, because the current zoom level exceeds the new max/min exceeds.
Leaving it in this state would mean that the engine is in an invalid state, so the zoom is adjusted to fit within the newly set bounds.

If you don't want to apply the bounds set by the user, you should wait before you set the user input in the `ZoomEngine`. I cannot think of a scenario where this wouldn't be sufficient.

konstant2016:
我目前遇到的问题是：根据子view的不同状态实时调节用户可以手动放大的最大级别，而这个调节对于用户来讲，是无感的，这就意味着在用户没有操作的情况下，显示画面上不可以自动调整大小，具体场景如下：
1，当数据处于A状态时，当前界面的缩放级别为2.0
2，当数据处于B状态时，当前界面的缩放级别为3.0
而在这两种状态切换时，除非用户手动触摸屏幕，否则不允许画面自动缩放
基于以上场景，我添加了scaleView参数，用来控制是否立刻缩放画面
如果源码中有更好的API可以实现上述需求，还请提示我一下，非常感谢～

markusressel:
Please use something like deepl.com to translate your message, since we cannot help you otherwise.

Automatic translation of the last message:
> The problem I am currently encountering is that the maximum level at which the user can manually zoom in is adjusted in real time according to the different states of the subview, and this adjustment is indifferent to the user, which means that no automatic resizing is possible on the display screen without user operation, and the specific scenarios are as follows.
1, when the data is in A state, the zoom level of the current interface is 2.0
2, when the data is in B state, the zoom level of the current interface is 3.0
And in these two states when switching, unless the user manually touch the screen, the screen does not allow automatic scaling
Based on the above scenario, I added the scaleView parameter to control whether the screen is scaled immediately or not
If there is a better API in the source code to achieve the above requirements, please let me know, thanks a lot~

I still don't understand the problem.

Given the zoom is currently at 2 and in state A,
When changing to state B, the zoom should not change, since 2 is still within the bounds of a max zoom of 3.
If the user now zooms to 2.5, and then switching to state A, the zoom level will be adjusted to 2, since the new maximum zoom level is 2 and 2.5 exceeds this level. Leaving it at 2.5 would result in an invalid state of the zoom engine, so we cannot just skip this.

If you only want to adjust zoom limits after user interaction, then simply wait for user interaction before applying the boundaries.

If you want to remember zoom level between states A and B, simply apply the zoom level in code when the user switches states, or touches the screen f.ex..

I don't see how it is beneficial or even necessary to leave the engine in an invalid state 🤔

konstant2016:
已经成功收到您的邮件

Issue #213, opened 2021-11-13
 Unwanted animated zoom after setMaxZoom call


sergsave:
### Describe the bug
ZoomLayout zooms when after `setMaxZoom(xx, ZoomApi.TYPE_REAL_ZOOM)` even current `realZoom` value less than `maxZoom` value
- Library version: 1.8

### To Reproduce
```
val zoomLayout = binding.zoomLayout

zoomLayout.realZoomTo(1.5f, false)

// Just wait while zoom applied
zoomLayout.postDelayed(1000) {
    // With current transformation the zoom value is greater than the realZoom value
    val zoom = zoomLayout.zoom // 2.21f
    val realZoom = zoomLayout.realZoom // 1.5f

    // After setMaxZoom call a layout automatically zooms with animation
    // But why? maxZoom value is less than current realZoom value
    val maxZoom = 2f
    zoomLayout.setMaxZoom(maxZoom, ZoomApi.TYPE_REAL_ZOOM)
}
```

### Expected behavior
The layout should not be zoomed when setting the maximum zoom, if the current zoom value is less than the setted maximum value.

sergsave:
Maybe this is because of this comparison
https://github.com/natario1/ZoomLayout/blob/94962c425e87a2014eb118c1811000495772cb38/library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt#L850

`zoom` has annotation `@Zoom`, `zoomManager.getMaxZoom()` has annotation `@ZoomApi.RealZoom`.
I think comparison of different types is not correct.

In `setMinZoom` value compares with `realZoom` value
https://github.com/natario1/ZoomLayout/blob/94962c425e87a2014eb118c1811000495772cb38/library/src/main/java/com/otaliastudios/zoom/ZoomEngine.kt#L885

PR #229, opened 2022-08-17
Fix setMaxZoom

quanlt:
It's related to [Unwanted animated zoom after setMaxZoom call #213](https://github.com/natario1/ZoomLayout/issues/213) issue.
I tried his solution and it worked. 
I am not sure why do you use zoom instead of realZoom, but replace it by realZoom fixed the animation issue.
This PR is a simple fix.


markusressel:
Haven't tested this myself, but the code makes sense to me.
Can we merge this @natario1 ?

markusressel (approved):

natario1:
Sure, makes sense!

